PR #315, opened 2022-05-02
development/filesync: introduce controller which sub-processes a server.

hugosantos:
This relies on the fact that foundation respects flag ordering. Node-based
flags are always appended to the list. So a language can set Command to
`filesync-controller`, and args to `[]string{root, fmt.Sprintf("%d", port),
originalCommand}` and we'll happily schedule the controller. Which then happily
starts the specified server, passing along any additional flags.

This is an alternative on the considered sidecar, as the sidecar requires a
shared volume to be set between the containers. And that would require copying
the root into the shared volume, adding non-trivial startup time.


n-g:
LGTM

PR #425, opened 2022-05-17
console: remember recent log sources

anikiej:
I moved err_context from tasks/ into console/ as I wanted to organize code related to `WithLogs` in the same place and the `tasks` package couldn't depend on consolesink.


n-g:
LGTM

PR #510, opened 2022-05-25
Split CodegenMultiError into a thread-safe error collector and an immutable CodegenMultiError

nichtverstehen:
This separates the concerns and avoids reclassifying the errors in CodegenMultiError at every read.


n-g:
LGTM

PR #564, opened 2022-06-07
std/grpc/httptranscoding: introduce package which represents the http/grpc transcoding.

hugosantos:
This is a step towards replacing the builtin gateway support within Go servers.
Depending on std/grpc/httptranscoding disables that builtin support. Although
this new package still does not yet configure the gateway.


n-g:
LGTM

PR #603, opened 2022-06-14
std/grpc/httptranscoding: compute domain names, rather than defer them to fn.

hugosantos:
This allows us to always produce fully formed resources, but most importantly
we're also in a position where we can configure Envoy with vhosts.

- std/grpc/httptranscoding: compute domain names, rather than defer them to fn.
- frontend/invocation: changed injection points to be typed, to support arguments in the future.


n-g:
LGTM

PR #867, opened 2022-09-01
internal/frontend: use Resources in ConfigurableVolume entries.

hugosantos:
This allows us to consistently make use of them when e.g. generating config images etc.


n-g:
LGTM

PR #938, opened 2022-10-26
Wire up diagnostics for service waiting

n-g:
The intention is to
a) display dial errors
b) forward diagnostics from pods that are supposed to back the service

Examples:

1) Forced failure through invalid address to dial
image

2) Strapi backend terminates prematurely due to invalid config
image

Fixes NSL-306


internal/runtime/kubernetes/kubeobserver/waiter.go
```diff
@@ -42,24 +41,23 @@ func WaitForCondition[Client any](ctx context.Context, cli Client, action *tasks
...
 		RuntimeSpecificHelp: fmt.Sprintf("kubectl -n %s describe %s %s", namespace, strings.ToLower(gvk.Kind), name),
 		Ready:               orchestration.Event_NOT_READY,
+		Scope:               scope,
 	}
 
 	switch {
 	case kubedef.IsGVKDeployment(gvk), kubedef.IsGVKStatefulSet(gvk), kubedef.IsGVKPod(gvk):
 		ev.Category = "Servers deployed"
+	case kubedef.IsGVKService(gvk):
+		ev.Category = "Services deployed"
```
hugosantos, line 56: Won't this add a considerable amount of output?

internal/runtime/kubernetes/kubeobserver/waiter.go
```diff
@@ -42,24 +41,23 @@ func WaitForCondition[Client any](ctx context.Context, cli Client, action *tasks
 	})
 }
 
-func PrepareEvent(gvk kubeschema.GroupVersionKind, namespace, name, desc string, deployable runtime.Deployable) *orchestration.Event {
+func PrepareEvent(gvk kubeschema.GroupVersionKind, namespace, name, desc string, scope string) *orchestration.Event {
```
hugosantos, line 44: Why change the signature? Isn't it clearer to either receive a deployable or not. Can still use this method to prepare an event for a non-deployable.

internal/planning/deploy/wait.go
```diff
@@ -164,19 +170,21 @@ func observeContainers(ctx context.Context, env cfg.Context, cluster runtime.Clu
...
-
-						var ctrs []*runtimepb.ContainerUnitWaitStatus
-						ctrs = append(ctrs, cws.Containers...)
-						ctrs = append(ctrs, cws.Initializers...)
-
-						for _, ctr := range ctrs {
-							if ctr.Status.Crashed || ctr.Status.Failed() {
-								failed = true
+						waitErr := &runtimepb.WaitError{}
+						if err := w.Opaque.UnmarshalTo(cws); err == nil {
```
hugosantos, line 174: This feels common enough that could be just lifted into Event_WaitStatus? I.e. a new field under it, rather than going through Opaque.

internal/runtime/kubernetes/kubeobserver/servicewaiter.go
```diff
@@ -59,13 +67,26 @@ func (w *serviceWaiter) Poll(ctx context.Context, c *k8s.Clientset) (bool, error
...
+
 	var count int
 	for _, port := range service.Spec.Ports {
-		addr := fmt.Sprintf("%s.%s.svc.cluster.local:%d", service.Name, service.Namespace, port.Port)
+		addr := fmt.Sprintf("%s.%s.svc.cluster.local:%d", w.name, w.namespace, port.Port)
 
 		rawConn, err := net.DialTimeout("tcp", addr, dialTimeout)
-		if err != nil {
-			fmt.Fprintf(console.Debug(ctx), "failed to dial %s: %v\n", addr, err)
+		if !w.isReady(pod.Items, err) {
```
hugosantos, line 89: I'm a bit confused about this. Ideally we wouldn't have to look for readiness at every step, but we would instead use the graph dependencies for that purpose. Why is checking for readiness required here?

internal/runtime/kubernetes/kubeobserver/waiter.go
n-g: This actually accomplishes that the output shrinks, as all services are displayed in a single block.

image

vs

image

This code does not decide whether to display the services in general. That happens as soon as they emit events.

hugosantos: Screen Shot 2022-11-02 at 16 03 11

I'm comparing with this, where there is no separate "Services deployed" block, apart from the service block at the end. Which I guess you don't subsume?

internal/runtime/kubernetes/kubeobserver/waiter.go
n-g: I can leave the signature as is and set the scope by hand later.

I mostly want to set a scope in general, since the fallback label that is printed is the resource ID, which is too k8s specific 

See:
image

internal/runtime/kubernetes/kubeobserver/servicewaiter.go
n-g: yeah, this is a bit messy. I mostly want to have a callback to emit events which can then trigger pod diagnostics if the service is not ready for a long time.
We could also just emit the dial error for now. It just felt like this will not be very helpful and us understanding which pods should serve this service, felt like a real benefit.

internal/runtime/kubernetes/kubeops/op_apply.go
```diff
@@ -262,14 +263,39 @@ func apply(ctx context.Context, desc string, scope []fnschema.PackageName, obj k
 			}}, nil
 
 	case kubedef.IsService(obj):
+		pkg, ok := obj.GetAnnotations()[kubedef.K8sServicePackageName]
+		if ok && !slices.Contains(scope, fnschema.PackageName(pkg)) {
+			// Refine scope with service package for understandability.
+			scope = append(scope, fnschema.PackageName(pkg))
+		}
```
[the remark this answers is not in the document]
n-g: SG!

internal/planning/deploy/wait.go
n-g: SG!

internal/runtime/kubernetes/kubeobserver/waiter.go
n-g: Good catch! I mostly tested with `ns test`

`ns deploy` indeed looks messy now.

image

PR #940, opened 2022-10-31
Implementing filesync for opaque dockerfiles.

nicolasalt:
Syntax:

```
mounts: "/app/src": {
	syncWorkspace: fromDir: "src"
}
```

Implementation:
If there is a "syncWorkspace" mount:
 - A layer with the "filesync-controller" is added to the built image.
 - `ns dev` uses a custom "hot reload" module that forwards filesync events to the dev observer (same way it works for node.js/web today).
 - The dev observer starts the filesync controller process inside the same container. When `ns dev` exists, it sends `Ctrl-C` to the controller process. As a failsafe, ns kills all controller processes before starting one.

The existing Node.js/Web filesync works the same way as before and not migrated to the new way in this PR.

### DIFF
--- a/internal/hotreload/hotreloadmodule.go
+++ b/internal/hotreload/hotreloadmodule.go
@@ -19,11 +19,10 @@ type hotReloadModule struct {
 	sink   wsremote.Sink
 }
 
-// If "triggerFullRebuildPredicate" returns true, a full rebuild will be triggered instead of a hot reload.
 func NewHotReloadModule(module build.Workspace, opts *languages.HotReloadOpts) build.Workspace {
 	return hotReloadModule{
 		module: module,
-		sink:   observerSink{opts.Sink, opts.TriggerFullRebuiltPredicate},
+		sink:   observerSink{opts.Sink, opts.EventProcessor},
 	}
 }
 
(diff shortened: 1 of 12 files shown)


nicolasalt:
* main
  * **PR #940** <a href="https://app.graphite.dev/github/pr/namespacelabs/foundation/940" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/namespacelabs/foundation/940?utm_source=stack-comment).

internal/hotreload/hotreloadmodule.go
```diff
@@ -40,15 +42,21 @@ func (w hotReloadModule) VersionedFS(rel string, observeChanges bool) compute.Co
...
 }
 
 func (obs observerSink) Deposit(ctx context.Context, events []*wscontents.FileEvent) (bool, error) {
+	processedEvents := []*wscontents.FileEvent{}
 	for _, ev := range events {
-		if obs.triggerFullRebuildPredicate != nil && obs.triggerFullRebuildPredicate(ev.Path) {
-			return false, nil
+		processedEvent := ev
+		if obs.eventProcessor != nil {
+			processedEvent = obs.eventProcessor(ev)
```
hugosantos, line 53: We need a comment here on why this API is necessary, for example: "Give an implementation the opportunity to rewrite the paths in event, to make them relative to a different directory."

I think we also need to have a signal on whether an event should skipped, or we should stop altogether. returning false here means stop all events.

I think something like the following is simpler:

```
if obs.eventProcessor == nil {
  return obs.sink.Deposit(ctx, events)
}

var transformed []*wscontents.FileEvent
for _, ev := range events {
  newEvent, continueProcessing := obs.eventProcessor(ev)
  if !continueProcessing {
    // processor signalled that no further events should be delivered.
    return false, nil
  }

  if newEvent != nil {
    transformed = append(transformed, newEvent)
  }
}

return obs.sink.Deposit(ctx, transformed)
```

nicolasalt: `whether an event should skipped, or we should stop altogether` is there a use-case for such an API at the moment?

hugosantos (approved):

PR #942, opened 2022-11-01
Key bindings: storing the "enabled" state inside the handler.

nicolasalt:


nicolasalt:
* main
  * **PR #942** <a href="https://app.graphite.dev/github/pr/namespacelabs/foundation/942" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/namespacelabs/foundation/942?utm_source=stack-comment).

hugosantos (commented):
The previous model was there to guarantee that the rendered state was always consistent with the keys being pressed. That's why there was a single authority that can change the sticky. And that authority pushes out updates (thus label being pure). To support multi-state I would have pushed the responsibility of handling the transition to the same place that handles a bool now. But if you feel this is a better design i'll have a deeper look.

nicolasalt:
I think the state management of components that can react to user input and display something on the screen (called "Handler"/"Keybinding" in this case) should be encapsulated inside these components. This model would be consistent with any other UI framework and allow to implement behavior of any complexity.
The layer(s) above should only propagate key events and update the "keybinding" UI when needed, ideally by listening to component updates but the current model of updating the UI after the synchronous component event handling also works up to a point.

PR #950, opened 2022-11-04
Pass telemetry headers when downloading ns packages.

nichtverstehen:


internal/artifacts/download/download.go
```diff
@@ -43,6 +63,7 @@ func (dl *downloadUrl) Action() *tasks.ActionEvent {
 
 func (dl *downloadUrl) Inputs() *compute.In {
 	inputs := compute.Inputs().Str("url", dl.url)
+	inputs = inputs.StrMap("headers", dl.headers)
```
hugosantos, line 66: In some way, this makes sense, but the implication is that the forwarding of telemetry will lead to that telemetry data being part of the cache key. I think it would be simpler now to have a boolean here that would not be added to the cache key (i.e. not part of inputs), that instructs ns to forward telemetry data. And that should be used with care as it is possible that a download may differ based on headers, so we should only use this option when we know that that's not the case (e.g. for nsboot downloads).

After that change feel free to submit.

nichtverstehen: Idea: remove both URL and headers from the Inputs whenever we have the expected hash. This way cache hits won't depend on how exactly we download the file. And the upside is that we won't need to make this part depend on telemetry. WDYT?

hugosantos:
@nichtverstehen I remember you mentioned that my change didn't cover everything you had in mind; will you come back to this or shall we close the PR?

PR #957, opened 2022-11-11
Enable env and arg to be set on tests.

n-g:
Fixes NSL-349, NSL-354

### DIFF
--- a/internal/frontend/cuefrontendopaque/tests.go
+++ b/internal/frontend/cuefrontendopaque/tests.go
@@ -8,6 +8,7 @@ import (
 	"context"
 
 	"namespacelabs.dev/foundation/internal/fnerrors"
+	"namespacelabs.dev/foundation/internal/frontend/cuefrontend/args"
 	"namespacelabs.dev/foundation/internal/frontend/cuefrontend/binary"
 	integrationparsing "namespacelabs.dev/foundation/internal/frontend/cuefrontend/integration/api"
 	"namespacelabs.dev/foundation/internal/frontend/fncue"
@@ -19,7 +20,9 @@ import (
 )
 
 type cueTest struct {
-	Servers []string `json:"serversUnderTest"`
+	Servers []string            `json:"serversUnderTest"`
+	Args    *args.ArgsListOrMap `json:"args"`
+	Env     *args.EnvMap        `json:"env"`
 }
 
 func parseTests(ctx context.Context, env *schema.Environment, pl parsing.EarlyPackageLoader, pkg *pkggraph.Package, v *fncue.CueV) ([]*schema.Test, error) {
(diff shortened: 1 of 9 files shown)


internal/parsing/env.go
```diff
@@ -2,24 +2,27 @@
...
 import (
 	"google.golang.org/protobuf/proto"
 	"namespacelabs.dev/foundation/internal/fnerrors"
 	"namespacelabs.dev/foundation/schema"
 )
 
-func SetEnv(merged []*schema.BinaryConfig_EnvEntry, entry *schema.BinaryConfig_EnvEntry) ([]*schema.BinaryConfig_EnvEntry, error) {
-	for _, existing := range merged {
-		if entry.Name == existing.Name {
+func MergeEnvs(base []*schema.BinaryConfig_EnvEntry, additional []*schema.BinaryConfig_EnvEntry) ([]*schema.BinaryConfig_EnvEntry, error) {
```
hugosantos, line 13: Add a test too? :-)

internal/frontend/cuefrontendopaque/tests.go
```diff
@@ -48,9 +51,20 @@ func parseTest(ctx context.Context, env *schema.Environment, pl parsing.EarlyPac
...
 
+	envVars, err := bits.Env.Parsed(pkg.Location.PackageName)
+	if err != nil {
+		return nil, err
+	}
+
 	out := &schema.Test{
 		Name:             name,
 		ServersUnderTest: bits.Servers,
+		Driver: &schema.Binary{
```
hugosantos, line 62: It feels confusing to have a Driver which is not an actual binary. I suggest just having a schema.BinaryConfig here, which then gets merged as you're doing below.

internal/support/env.go
```diff
@@ -0,0 +1,28 @@
...
+	"google.golang.org/protobuf/proto"
+	"namespacelabs.dev/foundation/internal/fnerrors"
+	"namespacelabs.dev/foundation/schema"
+)
+
+func MergeEnvs(base []*schema.BinaryConfig_EnvEntry, additional []*schema.BinaryConfig_EnvEntry) ([]*schema.BinaryConfig_EnvEntry, error) {
+	merged := append(base, additional...)
+
+	// Check for collisions.
+	cache := make(map[string]*schema.BinaryConfig_EnvEntry)
```
hugosantos, line 17: This is not a cache, it's an index. :-)

internal/support/env.go
```diff
@@ -0,0 +1,28 @@
...
+	for _, entry := range merged {
```
hugosantos, line 18: Isn't this the same implementation as parsing/env.go?

internal/testing/fixture.go
```diff
@@ -150,6 +150,7 @@ func PrepareTest(ctx context.Context, pl *parsing.PackageLoader, env cfg.Context
 		Command:            testBin.Command,
 		Args:               testBin.Args,
 		WorkingDir:         testBin.WorkingDir,
+		Env:                testDef.Driver.Config.Env,
```
hugosantos, line 153: This doesn't seem consistent; above we use testBin.Args, and here only Config.Env? Where's the merged ones?

internal/support/env.go
n-g: Was supposed to be a move.
Can't live here due to dep cycle. Cleaned up.

internal/testing/fixture.go
n-g: Fixed with https://github.com/namespacelabs/foundation/commit/5e91f09ab775eae835dbf4eac5dbbc9e1529dcdc
and aligned other places that did not propagate env.

gmichelo (approved):

PR #958, opened 2022-11-11
Orchestration: skip orchestrator deployment if the live image matches the pinned one.

n-g:


orchestration/prepare.go
```diff
@@ -96,3 +119,68 @@ func PrepareOrchestrator(ctx context.Context, targetEnv cfg.Configuration, clust
 
 	return &RemoteOrchestrator{cluster: boundCluster, server: focus.Proto(), endpoint: endpoint}, nil
 }
+
+func requiresUpdate(ctx context.Context, cluster runtime.Cluster, server planning.Server, prebuilts []*schema.Workspace_BinaryDigest) (bool, error) {
```
hugosantos, line 123: Since this method requires a planning.Server, we need to spend cycles doing package loading. Ideally this check runs before doing any package loading. You only need the server to get a proto, so you get the id. I think we can either (a) hardcode the id, or (b) have a separate label that is hardcoded.

orchestration/prepare.go
```diff
@@ -39,7 +50,12 @@ func RegisterPrepare() {
 }
 
 func PrepareOrchestrator(ctx context.Context, targetEnv cfg.Configuration, cluster runtime.Cluster, wait bool) (any, error) {
-	env, err := MakeOrchestratorContext(ctx, targetEnv)
+	prebuilts, err := fetchPrebuilts(ctx)
```
hugosantos, line 53: Thinking about this, this means that the CLI is blocked on an API call before it can check if the orchestrator is up to date. That both means that (a) we're spending anything between 50 and 1000ms on this call (it's not edge cached), and (b) we require being online to use ns.

The orchestrator can do this type of blocking work behind the scenes. What about we attempt to connect to the current orchestrator, and ask it if it is up to date. And the orchestrator could check the version from time to time in the background.

PR #965, opened 2022-11-20
resolve potential symlinks before initialize fsevents.EventStream

szferi:
following the fsnotify/fsevents recommendation, we should resolve potential symlinks with`filepath.EvalSymlinks` to get the target path to watch.


hugosantos:
Thank you for looking into this!

n-g (approved):

PR #967, opened 2022-11-21
internal/wscontents: ignore changes to files which are not valid paths for example includes '..'

szferi:
It prevents calling `checkChanges` on file paths that are invalid relative paths. For example, if the NS workspace is on $HOME and changes are made with vim, the `checkChanges` are called with ../.viminfo relative path which is an invalid path for the `fs.FS.Open` and it returns EINVAL that causes a fatal internal error. This is an attempt to fix these cases.

### DIFF
--- a/internal/wscontents/wscontents.go
+++ b/internal/wscontents/wscontents.go
@@ -12,6 +12,7 @@ import (
 	"io/fs"
 	"os"
 	"path/filepath"
+	"strings"
 	"time"
 
 	"github.com/dustin/go-humanize"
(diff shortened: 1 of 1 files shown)


internal/wscontents/wscontents.go
```diff
@@ -427,6 +427,12 @@ func handleEvents(ctx context.Context, debugLogger, userVisible io.Writer, absPa
 			return nil, false, fnerrors.InternalError("rel failed: %v", err)
 		}
 
+		// Ignore changes to files which are not valid paths for example includes ..
```
hugosantos, line 430: So if I understand correctly you see a path that is outside of `absPath`, which then leads to the resulting path being "../foobar" here? I.e. resolves to a relative path outside of absPath.

If that's the case, then I'd check for relative paths instead, e.g.

```
// Observed a change that is not within the destination we're observing. We'll ignore it.
if strings.HasPrefix(rel, "../") {
  fmt.Fprintf(..)
  continue
}
```

That would be slightly preferred vs fs.ValidPath, as ignoring other "invalid" paths may mask other bugs.

szferi: Ok. I updated the pull request.

n-g (approved):

PR #968, opened 2022-11-21
cmd/test: support full package references

szferi:


hugosantos:
So the interaction we're going for is enabling:

```
ns test golang/01-simple/server:startup-test
```

i.e. identifying a specific test (or tests). At the moment only the following is possible:

```
ns test golang/01-simple/server
```

Which will run all tests under `server`.

In order to support the case above, our location parsing needs to be aware of both packages ("golang/01-simple/server") and package references ("golang/01-simple/server:startup-test").

`fnfs.Location` represents a local filesystem location, that maps to a package. So that definition shouldn't change.

One option may be to extend `Locations` to have a `PackageRefs []*schema.PackageRef` constructed from arguments the user has provided. So while parsing arguments, if we find an argument that looks like a reference (i.e. has a `:` separator), then we mark it as a `PackageRef`.

Because not all commands support package references, this behavior should be opt-in. Similar to `RequireSingle` and `ReturnAllIfNoneSpecified`.

Then cmd/test can add any references that were parsed to its own list. And we can defer to its test loading code to do the final validation.

szferi:
The submitted code should implement your described behavior specifically to `ns test` but not change the argument parsing when we create `Locations`. Do you prefer to change the argument parsing still so other commands can support `PackageRefs` in the future?

hugosantos:
Sorry for the delay! Yeap -- it would be best to change Locations so we don't have to change the meaning of `Location`. Thank you for the help here.

szferi:
Ok, it took me some experimentation to find the right place to handle the parsing according to your suggestion because I tried to limit the scope of the change, but I ended up pushing down closest to the argument parsing. That required the change to be a bit more involved. Let me know what you think.

PR #972, opened 2022-11-23
Consistently using BinaryConfig everywhere.

nicolasalt:
Moving "command" to the root level of the "server" syntax.


nicolasalt:
* main
  * **PR #972** <a href="https://app.graphite.dev/github/pr/namespacelabs/foundation/972" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/namespacelabs/foundation/972?utm_source=stack-comment).

internal/integrations/opaque/opaque.go
```diff
@@ -85,12 +85,9 @@ func (OpaqueIntegration) PrepareRun(ctx context.Context, server planning.Server,
...
 		}
 
-		config := binary.Config
-		if config != nil {
-			run.WorkingDir = config.WorkingDir
-			run.Command = config.Command
-			run.Args = config.Args
-			run.Env = config.Env
+		err = run.BinaryConfig.MergeFrom(binary.Config)
+		if err != nil {
```
hugosantos, line 89: Nit: please scope err, rather than re-using variables. It makes these types of declarations simpler to maintain:

```
if err := run.BinaryConfig.MergeFrom(...); err != nil {
```

internal/integrations/opaque/opaque.go
```diff
@@ -85,12 +85,9 @@ func (OpaqueIntegration) PrepareRun(ctx context.Context, server planning.Server,
 			return err
 		}
 
-		config := binary.Config
-		if config != nil {
-			run.WorkingDir = config.WorkingDir
-			run.Command = config.Command
-			run.Args = config.Args
-			run.Env = config.Env
+		err = run.BinaryConfig.MergeFrom(binary.Config)
```
hugosantos, line 88: Leaving our merge semantics to proto merge semantics feels like a future problem. Lets define our own merge function, with clear documentation of what gets replaced by what.

internal/planning/deploy/deploy.go
```diff
@@ -782,12 +779,10 @@ func prepareRunOpts(ctx context.Context, stack *planning.Stack, srv planning.Ser
 		return err
 	}
 
-	if merged.WorkingDir != "" {
-		out.MainContainer.WorkingDir = merged.WorkingDir
+	err = out.MainContainer.BinaryConfig.MergeFrom(merged)
+	if err != nil {
```
hugosantos, line 783: Same here, (a) lets define our own merge, and (b) avoid re-using variables in Go (can easily use if statement scoping).

internal/planning/deploy/deploy.go
```diff
@@ -782,12 +779,10 @@ func prepareRunOpts(ctx context.Context, stack *planning.Stack, srv planning.Ser
...
+		return err
 	}
-	out.MainContainer.Args = append(out.MainContainer.Args, merged.Args...)
-	out.MainContainer.Env = append(out.MainContainer.Env, srv.Proto().MainContainer.Env...)
-	out.MainContainer.Env = append(out.MainContainer.Env, merged.Env...)
```
hugosantos, line 790: Where do these merges occur now? In the call above? Does that mean that there are more fields being changed, not just Args and Env?

internal/planning/invocation/invocation.go
```diff
@@ -33,11 +32,8 @@ type Invocation struct {
 	Image                compute.Computable[oci.ResolvableImage]
 	PublicImageID        *oci.ImageID
 	SupportedToolVersion int
-	Command              []string
-	Args                 []string
-	Env                  []*schema.BinaryConfig_EnvEntry
+	BinaryConfig         *schema.BinaryConfig
```
hugosantos, line 35: This feels like a step backwards because we're now embedding a single definition into domain-specific invocations. Previously, if one would add a new field to BinaryConfig, it would be clear whether there's a place for it in e.g. `Invocation` or not -- now there's an assumption that BinaryConfig is universally applicable, and that doesn't feel future-proof.

internal/runtime/docker/invoke.go
```diff
@@ -58,11 +58,10 @@ func (r ToolRuntime) RunWithOpts(ctx context.Context, opts rtypes.RunToolOpts, o
 
 	return tasks.Action("container.execute").
 		LogLevel(2).
-		Arg("command", opts.Command).
+		Arg("binaryConfig", opts.BinaryConfig).
```
hugosantos, line 61: This means that a whole json is being printed out, that doesn't feel quite right.

internal/runtime/interface.go
```diff
@@ -289,11 +289,8 @@ type GroundedSecret struct {
 }
 
 type ContainerRunOpts struct {
-	WorkingDir         string
 	Image              oci.ImageID
-	Command            []string
-	Args               []string
-	Env                []*schema.BinaryConfig_EnvEntry
+	BinaryConfig       *schema.BinaryConfig
```
hugosantos, line 293: I mentioned this above on `Invocation` but it's the same problem here -- it doesn't feel like this is going in the right direction. Previously ContainerRunOpts' interface was fairly well defined. Although it was already a challenge that it re-used BinaryConfig_EnvEntry, which has multiple meanings.

Embedding BinaryConfig means that future updates to BinaryConfig will end up being part of the runtime interface, and that's not quite right, as we're mixing two different APIs. I think it would be best to keep the fields that a runtime interface can act on explicitly declared, so Args, Command, Env etc separate as before.

internal/runtime/kubernetes/deployment.go
```diff
@@ -30,7 +30,6 @@ import (
 	"namespacelabs.dev/foundation/internal/fnerrors"
```
hugosantos, line 30: The fact that runtime/kubernetes/deployment needs to be touched as part of this change doesn't seem right to me -- BinaryConfig should have been exclusively about parsing/early planning.

schema/binary.go
```diff
@@ -0,0 +1,57 @@
...
+
+package schema
+
+import (
+	"google.golang.org/protobuf/proto"
+	"namespacelabs.dev/foundation/internal/fnerrors"
+)
+
+// Modifies the instance in place.
+func (bc *BinaryConfig) MergeFrom(other *BinaryConfig) error {
```
hugosantos, line 13: I'd like to avoid OO for protos. Protos should be exclusively data, and attaching behavior to them makes their API surface cover two areas, and makes it more difficult to reason about e.g. versioning.

We have a few existing helper methods for protos, that cast to domain-specific types (e.g. string to PackageName) which are OK because they don't really implement any other behavior beyond type casting.

You'll see from my comments above that I got confused by the use of `MergeFrom` as I assumed you were using proto's Merge methods.

Lets keep these type of methods in a package that makes sense for them, so its clearly readable what the intended behavior is, e.g. `deploy.CombineBinaryConfig()`, or `planning.CombineBinaryConfig`.

schema/binary.go
```diff
@@ -0,0 +1,57 @@
...
+	// TODO: consider if this should be a replacement.
+	bc.Args = append(bc.Args, other.Args...)
+
+	var err error
+	bc.Env, err = mergeEnvs(bc.Env, other.Env)
+	return err
+}
+
+// Returns a deep copy with args added to the end of the list.
+func (bc *BinaryConfig) WithArgs(args ...string) *BinaryConfig {
```
hugosantos, line 34: I see the usefulness of this method, but similarly I think its best to keep it out of protos.

internal/integrations/opaque/opaque.go
hugosantos: (I see now that you've added these methods to the protos -- lets not do that, see my comment below regarding OOP).

nicolasalt:
I removed BinaryConfig for Invocation, ContainerOpts, etc., and kept it only in parsing.
"merge" functions are not needed anymore since merge now happens between BinaryConfig and individual fields on other protos.

PTAL.

PR #975, opened 2022-11-24
implements ability to create new environment with prepare

szferi:
The implementation supports environment purpose, runtime, and labels:
For example:
```
ns prepare local --env=foobar --create-env \
      --create-env-args='purpose=development,runtime=kubernetes,"labels=a=b,c=d"'
```


hugosantos (commented):
Thank you for this. Left two comments, but otherwise looks good to me!

szferi:
I updated the PR following your feedback

PR #981, opened 2022-11-28
cli: add `mod init` command that creates an initial ns-workspace.cue.

szferi:


internal/cli/cmd/mod.go
```diff
@@ -22,12 +31,54 @@ func NewModCmd() *cobra.Command {
...
+					MinimumApi: versions.MinimumAPIVersion,
+				},
+			}
+			mod, err := parsing.NewModule(ctx, dir, cuefrontend.WorkspaceFile, w)
+			if err != nil {
+				return err
+			}
+			root := parsing.NewRoot(mod, mod)
+			err = fnfs.WriteWorkspaceFile(ctx, nil, root.ReadWriteFS(), mod.DefinitionFile(), func(w io.Writer) error {
+				return mod.FormatTo(w)
```
n-g, line 69: Nit: we typically move the assignment into the `if` block if only err is returned.
E.g.

```
if err := fnfs.WriteWorkspaceFile(ctx, nil, root.ReadWriteFS(), mod.DefinitionFile(), func(w io.Writer) error {
	return mod.FormatTo(w)
}); err != nil {
```

internal/cli/cmd/mod.go
```diff
@@ -22,12 +31,54 @@ func NewModCmd() *cobra.Command {
...
+				EnvSpec:    cfg.DefaultWorkspaceEnvironmets,
+				Foundation: &schema.Workspace_FoundationRequirements{
+					MinimumApi: versions.MinimumAPIVersion,
+				},
+			}
+			mod, err := parsing.NewModule(ctx, dir, cuefrontend.WorkspaceFile, w)
+			if err != nil {
+				return err
+			}
+			root := parsing.NewRoot(mod, mod)
```
n-g, line 67: Nit: Since `dir` is already absolute, you can skip `root` and just pass `fnfs.ReadWriteLocalFS(dir)` below.

internal/frontend/cuefrontend/module.go
```diff
@@ -85,6 +86,43 @@ func moduleFrom(ctx context.Context, dir, workspaceFile string, data []byte) (pk
...
 }
 
+func (moduleLoader) NewModule(ctx context.Context, dir, workspaceFile string, w *schema.Workspace) (pkggraph.WorkspaceData, error) {
+	val, err := decodeWorkspace(w)
+	if err != nil {
+		return nil, err
+	}
+	return workspaceData{
+		absPath:        dir,
+		definitionFile: workspaceFile,
```
n-g, line 96: New modules should always use the new filename. Let's pin `cuefrontend.WorkspaceFile` here and remove the parameter.

internal/cli/cmd/mod.go
```diff
@@ -22,12 +31,54 @@ func NewModCmd() *cobra.Command {
...
+			}
+			fmt.Println("Creating initial workspace.")
+			w := &schema.Workspace{
+				ModuleName: args[0],
+				EnvSpec:    cfg.DefaultWorkspaceEnvironmets,
+				Foundation: &schema.Workspace_FoundationRequirements{
+					MinimumApi: versions.MinimumAPIVersion,
+				},
+			}
+			mod, err := parsing.NewModule(ctx, dir, cuefrontend.WorkspaceFile, w)
```
n-g, line 63: I'm wondering if this is overkill and we should just use a simple text template.

How fast is `ns mod init` right now? 
I'd expect it to be feel instant. If it does, let's roll with this approach.
Otherwise, let's speed it up and just use a text template :)

n-g (changes requested):

internal/cli/cmd/mod.go
```diff
@@ -22,12 +31,54 @@ func NewModCmd() *cobra.Command {
...
+			}
+			root := parsing.NewRoot(mod, mod)
+			err = fnfs.WriteWorkspaceFile(ctx, nil, root.ReadWriteFS(), mod.DefinitionFile(), func(w io.Writer) error {
+				return mod.FormatTo(w)
+			})
+			if err != nil {
+				return err
+			}
+			fmt.Println("Running 'ns mod tidy' command to finalize the workspace.")
+			RunCommand(ctx, []string{"mod", "tidy"})
```
n-g, line 75: `return RunCommand(...`

n-g (approved):

PR #1013, opened 2022-12-18
library/oss/localstack: allow redefine server specification similar to postgres using template

szferi:
I followed the example of the Postgres database and cluster models. However, with this change, the S3:Bucket resource with the Localstack provider needs to have a `server` resource defined even if we do not want to overwrite anything. This is the same with Postgres Database and Postgres Cluster. Unfortunately, I could not figure out how to specify defaults for this. 
As per usual, naming is hard. There are a lot of things called "Server" in this PR. I welcome a better name suggestion.

### DIFF
--- a/internal/testdata/integrations/dockerfile/complex/server.cue
+++ b/internal/testdata/integrations/dockerfile/complex/server.cue
@@ -11,7 +11,9 @@ server: {
 			resource: "namespacelabs.dev/foundation/internal/testdata/integrations/resources/instances:test1"
 			fieldRef: "url"
 		}
-		INGRESS: fromServiceIngress: ":webapi"
+		if $env.purpose != "PRODUCTION" {
+			INGRESS: fromServiceIngress: ":webapi"
+		}
 	}
 
 	services: {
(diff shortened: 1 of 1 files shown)


library/oss/localstack/providers.cue
```diff
@@ -1,12 +1,28 @@
...
-			server: {
-				class:  "namespacelabs.dev/foundation/library/runtime:Server"
-				intent: "namespacelabs.dev/foundation/library/oss/localstack/server"
-			}
+		inputs: {
+			// TODO define default
+			"server": "namespacelabs.dev/foundation/library/oss/localstack:Server"
 		}
 	}
+	"namespacelabs.dev/foundation/library/oss/localstack:Server": {
```
hugosantos, line 10: I suggest we use "Cluster" to keep some parity with postgres. Although it's not 100% accurate, that's kind of what we're trying to do here.

hugosantos (approved):
Apart from a small rename from Server->Cluster, looks good to me. thanks!

library/oss/localstack/providers.cue
szferi: I renamed the Server to Cluster and added default with last commit

PR #1031, opened 2022-12-21
[WIP] Controller skeleton + CRD definition

gmichelo:


gmichelo:
TODO list (not in order of priority):

- [ ] Autogenerate CRD deepcopy.go files and YAML manifests, as one would do with kubebuilder ([example of makefile](https://github.com/gmichelo/k8s-network-prober-operator/blob/main/Makefile)). As of now, CRD YAML manifest and deepcopy functions are manually written.
- [ ] Access to image registry is set insecure to make it work with local image registry. Find proper fix.
- [ ] Image link must use the local image registry's internal port (i.e. 5000) rather than externally exposed port (e.g. 57321). Possible solution: on-the-fly automatic rewrites?
- [ ] Implement undeploy/cancellation use-case. Orchestrator gRPC server needs to implement an `undeploy` method.
- [ ] Set ownership-ref for K8s objects created by Orchestrator. 
- [ ] Properties in `Revision.Status`? User might want to know the deployment ID to refer to it later
- [ ] Set events on `Revision` object
- [ ] Add code comments
- [ ] User documentation

PR #1032, opened 2022-12-22
chore: deleting the unused AWS S3 extension.

nicolasalt:


nicolasalt:
* main
  * **PR #1032** <a href="https://app.graphite.dev/github/pr/namespacelabs/foundation/1032" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/namespacelabs/foundation/1032?utm_source=stack-comment).

universe/aws/client/factory.go
```diff
@@ -18,23 +17,12 @@ import (
...
-			return aws.Config{}, errors.New("when running without universe/aws/irsa, aws credentials are required to be set")
-		}
-
-		core.Log.Printf("[aws/client] using shared credentials at %q", cf.SharedCredentialsPath)
-
-		optFns = append(optFns, config.WithSharedCredentialsFiles([]string{cf.SharedCredentialsPath}))
-	} else {
-		core.Log.Printf("[aws/client] using web identity credentials at %q", tokenFile)
-	}
+	tokenFile := os.Getenv("AWS_WEB_IDENTITY_TOKEN_FILE")
```
hugosantos, line 24: The error path from above should be maintained, i.e. if the env var is not set, then we should fail.

universe/storage/s3/wire.go
```diff
@@ -97,32 +97,29 @@ func createClient(ctx context.Context, factory client.ClientFactory, minioCreds
...
-
-		resolver := aws.EndpointResolverWithOptionsFunc(func(service, region string, options ...interface{}) (aws.Endpoint, error) {
-			return aws.Endpoint{
-				PartitionID:   "aws",
-				URL:           *minioEndpoint,
-				SigningRegion: region,
-			}, nil
-		})
+	if *minioEndpoint == "" {
+		return nil, fmt.Errorf("aws S3 is not supported")
```
hugosantos, line 101: Aren't you removing the irsa path here as well? I.e. when `AWS_WEB_IDENTITY_TOKEN_FILE` is set. This will break internal, where we use s3.

hugosantos (changes requested):

universe/aws/client/factory.go
nicolasalt: Done.

universe/storage/s3/wire.go
nicolasalt: Reverted, I initially started to delete more than needed but then forgot to revert this path.

universe/storage/s3/internal/prepare/prepares3.go
```diff
@@ -227,11 +221,7 @@ func (provisionHook) Apply(ctx context.Context, req provisioning.StackRequest, o
 			&schema.BinaryConfig_EnvEntry{Name: "MINIO_PASSWORD", FromSecretRef: schema.MakePackageRef("namespacelabs.dev/foundation/universe/storage/minio/creds", "root-password")})
 		commonArgs = append(commonArgs, fmt.Sprintf("--%s=%s", useMinioFlag, service))
 	} else {
-		for _, secret := range col.SecretsOf("namespacelabs.dev/foundation/universe/aws/client") {
-			if secret.Name == "aws_credentials_file" {
-				initArgs = append(initArgs, fmt.Sprintf("--aws_credentials_file=%s", secret.FromPath))
-			}
-		}
+		return fmt.Errorf("AWS S3 is not supported")
```
hugosantos, line 224: I think you should just remove this path; this is only trying to forward --aws_credentials_file if its set. No forward is done anymore, but AWS S3 is still supported, but requires irsa.

hugosantos (changes requested):

universe/storage/s3/internal/prepare/prepares3.go
nicolasalt: Done.

hugosantos (approved):

PR #1043, opened 2023-01-10
Cloud build support

szferi:
On cloud build, mounting the $HOME/.config/gcloud does not work. Instead, one should attach `cloudbuild` network to the container to make the config-helper work (see:  https://cloud.google.com/build/docs/build-config-file-schema#network )


szferi:
The added `--gcloud_use_host_binary` flags solve this problem better.

PR #1045, opened 2023-01-11
dash not allowed in the database name either

szferi:


library/oss/postgres/prepare/database/main.go
```diff
@@ -73,7 +73,7 @@ func ensureDatabase(ctx context.Context, cluster *postgresclass.ClusterInstance,
 		// https://www.postgresql.org/docs/9.5/xfunc-sql.html
 		// `existsDb` already uses the database name as an SQL argument, so we already passed its validation.
 		// Still, let's do some basic sanity checking (whitespaces are forbidden), as we need to use Sprintf here.
-		if len(strings.Fields(name)) > 1 {
+		if len(strings.Fields(name)) > 1 || strings.Contains(name, "-") {
```
n-g, line 76: The original intention of this check was to provide some basic SQL injection protection (though existsDatabase would already fail).

Is the intention of the addition to provide more understandable errors?
I agree that `ERROR: syntax error at or near "-" (SQLSTATE 42601)` is not very ideal.
I had a look at https://www.postgresql.org/docs/current/sql-syntax-lexical.html#SQL-SYNTAX-IDENTIFIERS for what it would take to fully validate database names. 

> SQL identifiers and key words must begin with a letter (a-z, but also letters with diacritical marks and non-Latin letters) or an underscore (_). Subsequent characters in an identifier or key word can be letters, underscores, digits (0-9), or dollar signs ($).

Sounds rather complicated, so I'm happy to just go with this simple check now. I'll add a reference to the doc.

n-g (approved):

PR #1047, opened 2023-01-11
internal/fnapi: request a JWT token from Github OIDC provider

edganiukov:
This change adds an option to obtain JWT token from Github OIDC provider.
By default it requests a token only when executed in Github Actions environment.


internal/github/jwt.go
```diff
@@ -0,0 +1,69 @@
...
+
+	resp, err := http.DefaultClient.Do(req)
+	if err != nil {
+		return "", fnerrors.InvocationError("jwt", "failed to get github JWT: %w", err)
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK {
+		if resp.StatusCode == http.StatusUnauthorized {
+			return "", auth.ErrRelogin
```
n-g, line 53: This error is not ideal here.

If we get and unauth error while fetching the Github JWT, running `ns login` would not help resolve it.

internal/fnapi/userauth.go
```diff
@@ -29,15 +30,24 @@ type UserAuth struct {
 }
 
 func (user UserAuth) GenerateToken(ctx context.Context) (string, error) {
-	if user.Clerk != nil {
+	switch {
+	case os.Getenv("GITHUB_ACTIONS") == "true":
+		jwt, err := github.JWT(ctx, "")
```
n-g, line 35: Let's add a parameter comment - or remove the parameter until we need it.

n-g (approved):

internal/github/jwt.go
edganiukov: Added a comment (I am just not sure if default one is OK, most likely we would need to set one)

internal/fnapi/userauth.go
edganiukov: Added a comment (I agree not nice to have just empty string there, but most likely we would need it, not sure tho)

internal/github/jwt.go
edganiukov: dropped this part and return a general error

internal/fnapi/userauth.go
```diff
@@ -29,15 +30,27 @@ type UserAuth struct {
...
+
+	case os.Getenv("GITHUB_ACTIONS") == "true":
+		jwt, err := github.JWT(ctx, "") // XXX: do not set "audience" and use default one
+		if err != nil {
+			return "", err
+		}
+		return fmt.Sprintf("gh-jwt:%s", jwt), nil
+
+	default:
+		return "", fnerrors.New("no authentication method available")
```
n-g, line 52: We can actually return auth.ErrRelogin here, since ns login will populate user.InternalOpaque and hence avoid this error.

edganiukov: done

PR #1070, opened 2023-01-30
server def supports host networking

htr:


internal/planning/deploy/deploy.go
```diff
@@ -885,6 +885,7 @@ func prepareRunOpts(ctx context.Context, stack *planning.Stack, srv planning.Ser
 	out.MainContainer.Env = append(out.MainContainer.Env, main.Env...)
 	out.MainContainer.Env = append(out.MainContainer.Env, merged.Env...)
 	out.MainContainer.Privileged = main.GetSecurity().GetPrivileged()
+	out.MainContainer.HostNetwork = main.GetSecurity().GetHostNetwork()
```
gmichelo, line 888: No DNS config here?

htr: No need at this stage. I only set it when creating the spec because the default isn't appropriate for hostNetworking.

hugosantos (approved):

PR #1081, opened 2023-01-30
internal/providers/nscloud: use tenant tokens for image registry authN

edganiukov:
Fixes NSL-537

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -144,9 +144,8 @@ jobs:
       - name: Make ns executable
         run: chmod +x ns && mv ns /tmp/ns
 
-      - name: Login
-        # Consider using secrets.GITHUB_TOKEN
-        run: echo ${{ secrets.NSL_GH_SERVICE }} | /tmp/ns login robot github.com/namespacelabs/foundation
+      - name: Exchange Github token
+        run: /tmp/ns auth exchange-github-token
 
       - name: Run tests
         # Consider removing --also_report_start_events
@@ -200,9 +199,8 @@ jobs:
       - name: Make ns executable
         run: chmod +x ns && mv ns /tmp/ns
 
-      - name: Login
-        # Consider using secrets.GITHUB_TOKEN
-        run: echo ${{ secrets.NSL_GH_SERVICE }} | /tmp/ns login robot github.com/namespacelabs/foundation
+      - name: Exchange Github token
+        run: /tmp/ns auth exchange-github-token
 
       - name: Configure AWS credentials
         uses: aws-actions/configure-aws-credentials@v1
@@ -280,9 +278,8 @@ jobs:
       - name: Make ns executable
         run: chmod +x ns && mv ns /tmp/ns
 
-      - name: Login
-        # Consider using secrets.GITHUB_TOKEN
-        run: echo ${{ secrets.NSL_GH_SERVICE }} | /tmp/ns login robot github.com/namespacelabs/${{ matrix.repo }}
+      - name: Exchange Github token
+        run: /tmp/ns auth exchange-github-token
 
       - name: Run tests
         run: |
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -199,8 +199,9 @@ jobs:
       - name: Make ns executable
         run: chmod +x ns && mv ns /tmp/ns
 
-      - name: Exchange Github token
-        run: /tmp/ns auth exchange-github-token
+      - name: Login
+        # Consider using secrets.GITHUB_TOKEN
+        run: echo ${{ secrets.NSL_GH_SERVICE }} | /tmp/ns login robot github.com/namespacelabs/foundation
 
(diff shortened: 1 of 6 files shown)


internal/providers/nscloud/registry.go
```diff
@@ -7,36 +7,36 @@ package nscloud
...
 
 type nscloudRegistry struct {
 	clusterID string
 	registry  *api.ImageRegistry
 }
 
-const registryAddr = "registry-fgfo23t6gn9jd834s36g.prod-metal.namespacelabs.nscloud.dev"
-
-func RegisterRegistry() {
+func RegisterRegistry(imageRegistry *api.ImageRegistry) error {
```
n-g, line 39: Hmm, passing the registry here feels odd, as we need to call this from many places then (as you do).
I would prefer, if we register a factory for the keychain that dynamically fetches the creds from checking the cluster ID, as we do for the registry.

From what I can tell, this would require a larger refactor to how the keychain is handled (making `cfg.Configuration` available) so let's check with @hugosantos before going down this path.

internal/providers/nscloud/registry.go
```diff
@@ -106,53 +111,83 @@ func (dk defaultKeychain) Resolve(ctx context.Context, r authn.Resource) (authn.
 	if !strings.HasSuffix(r.RegistryStr(), ".nscluster.cloud") {
 		return authn.Anonymous, nil
 	}
-
-	ref, err := name.ParseReference(r.String())
+	creds, err := loadRegistryConfig(r.RegistryStr())
 	if err != nil {
 		return nil, err
```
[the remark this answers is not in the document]
edganiukov: Hm, but after cluster creation, `ns` exists therefore I need to store somewhere user/pass.

internal/providers/nscloud/registry.go
hugosantos: As @n-g says, this information should be tied to the cluster, and should be looked up dynamically during runtime. And in fact never stored. If a user re-uses a nscloud cluster, this information will be fetched again from the VM service.

edganiukov: Changed to request scopes tenant token to access registry.

.github/workflows/e2e.yaml
```diff
@@ -200,9 +199,8 @@ jobs:
       - name: Make ns executable
         run: chmod +x ns && mv ns /tmp/ns
 
-      - name: Login
```
n-g, line 203: Let's keep the old login for `ns deploy --env=staging`

internal/providers/nscloud/registry.go
```diff
@@ -107,52 +97,42 @@ func (dk defaultKeychain) Resolve(ctx context.Context, r authn.Resource) (authn.
...
-	req.Header.Add("Authorization", "Bearer "+tok)
+		if !os.IsNotExist(err) {
+			return "", err
+		}
 
-	resp, err := http.DefaultClient.Do(req)
-	if err != nil {
-		return nil, err
-	}
+		// TODO exchange user auth for tenant tokens.
```
n-g, line 115: Aren't you already doing this with `fnapi.ExchangeUserToken`?

internal/providers/nscloud/registry.go
```diff
@@ -107,52 +97,42 @@ func (dk defaultKeychain) Resolve(ctx context.Context, r authn.Resource) (authn.
...
-
-	values := url.Values{}
-	values.Add("scope", fmt.Sprintf("repository:%s:push,pull", ref.Context().RepositoryStr()))
-	values.Add("service", "Authentication")
-
-	req, err := http.NewRequestWithContext(ctx, "GET", fmt.Sprintf("https://%s?%s", loginEndpoint, values.Encode()), nil)
-	if err != nil {
-		return nil, err
-	}
+	exchangeToken := func(ctx context.Context) (string, error) {
```
n-g, line 100: Shall we define this in a shareable place so that we can also use it from https://github.com/namespacelabs/foundation/blob/7deff0f173d3a3c1ddc5606d649b8ebf2135e8c8/internal/providers/nscloud/api/rpc.go#L37-L38 ?

n-g (approved):

.github/workflows/e2e.yaml
edganiukov: True, reverted.

internal/providers/nscloud/registry.go
edganiukov: Left over from old way. Dropped.

internal/providers/nscloud/registry.go
edganiukov: Hm, I was thinking to move it to `auth` pkg, but then there will be an import cycle. What would be the better place? `interna;/fnapi`, `internal/providers/nscloud/api`?

n-g: I'd start with `internal/providers/nscloud/api` since these tokens are nscloud specific.

edganiukov: done

PR #1083, opened 2023-01-31
nscloud: Implement flag to wait for kube system resources.

n-g:
Fixes NSL-515

### DIFF
--- a/internal/providers/nscloud/wait.go
+++ b/internal/providers/nscloud/wait.go
@@ -23,11 +23,12 @@ import (
 	"namespacelabs.dev/foundation/std/tasks"
 )
 
-const (
-	kubeSystem       = "kube-system"
-	coredns          = "coredns"
-	localProvisioner = "local-path-provisioner"
-)
+const kubeSystem = "kube-system"
+
+var deployments = map[string]struct{}{
+	"coredns":                {},
+	"local-path-provisioner": {},
+}
 
 func WaitKubeSystem(ctx context.Context, cluster *api.KubernetesCluster) error {
 	return tasks.Action("cluster.wait-kube-system").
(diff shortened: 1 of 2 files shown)


internal/providers/nscloud/wait.go
```diff
@@ -0,0 +1,97 @@
...
+	"namespacelabs.dev/foundation/internal/fnerrors"
+	"namespacelabs.dev/foundation/internal/providers/nscloud/api"
+	"namespacelabs.dev/foundation/internal/runtime/kubernetes/client"
+	"namespacelabs.dev/foundation/internal/runtime/kubernetes/kubeobserver"
+	"namespacelabs.dev/foundation/std/tasks"
+)
+
+const (
+	kubeSystem       = "kube-system"
+	coredns          = "coredns"
```
hugosantos, line 28: Make this a string slice instead?

internal/providers/nscloud/wait.go
```diff
@@ -0,0 +1,97 @@
...
+			return true, nil
+		}); err != nil {
+			return fnerrors.New("failed to wait for deployments in namespace %q: %w", kubeSystem, err)
+		}
+
+		fmt.Fprintf(console.Debug(ctx), "got %d kube-system deployments\n", len(list.Items))
+
+		eg := executor.New(ctx, "wait")
+
+		for _, d := range list.Items {
```
hugosantos, line 76: This waits for everything in kube-system, doesn't seem quite right. Should we not wait only for the deployments listed above?

n-g: It felt more conservative to me to wait for all of kube-system.
In practice, I've only seen it wait for the two. Updated as you believe they suffice :)

gmichelo (approved):

PR #1088, opened 2023-02-02
Introduce minimal nsc command

n-g:
Release it along with ns releases.
Do not perform version checks yet.

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -1,7 +1,7 @@
 project_name: ns
 
 builds:
-  - id: linux
+  - id: linux-ns
     env:
       - CGO_ENABLED=0
     main: ./cmd/ns
@@ -16,7 +16,7 @@ builds:
       - amd64
       - arm64
 
-  - id: macos
+  - id: macos-ns
     env:
       - CGO_ENABLED=1
     main: ./cmd/ns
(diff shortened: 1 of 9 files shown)


.goreleaser.yaml
```diff
@@ -31,6 +31,49 @@ builds:
...
+      - -s -w -X namespacelabs.dev/foundation/internal/cli/version.Tag=v{{.Version}}
+    goos:
+      - linux
+    goarch:
+      - amd64
+      - arm64
+
+  - id: macos-nsc
+    env:
+      - CGO_ENABLED=1
```
hugosantos, line 51: Ideally we don't need CGO_ENABLE for nsc as it's only needed for fsevents, which we shouldn't be pulling in.

internal/cli/version/version.go
```diff
@@ -67,7 +67,12 @@ func isDevelopmentBuild(ver *storage.NamespaceBinaryVersion) bool {
 	return ver.BuildTime == nil || ver.Version == DevelopmentBuildVersion
 }
 
-func ShouldCheckUpdate(ver *storage.NamespaceBinaryVersion) bool {
+func ShouldCheckUpdate(name string, ver *storage.NamespaceBinaryVersion) bool {
+	if name != "ns" {
```
hugosantos, line 71: Hm, this is a bit clunky. :D Could this be a flag in DoMain()?

gmichelo (approved):

PR #1090, opened 2023-02-03
allows user defined ingress annotations

szferi:
Extend the ingress definitions with user-defined annotations so she can do something like this:
```
ingress: {
	provider: "github.com/ns-test/cmingress"
	domains: ["example.com", "www.example.com"]
	annotations: {
		"cert-manager.io/cluster-issuer": "cert-issuer-staging"
	}
}
```


internal/frontend/cuefrontendopaque/service.go
```diff
@@ -46,10 +46,12 @@ type cueIngress struct {
...
 type CueIngressDetails struct {
 	// Key is domain.
-	HttpRoutes     map[string][]string `json:"httpRoutes"`
-	ProviderClass  string              `json:"provider"`
-	Domains        []string            `json:"domains"`
-	AllowedOrigins []string            `json:"allowed_origins"`
+	HttpRoutes    map[string][]string `json:"httpRoutes"`
+	ProviderClass string              `json:"provider"`
+	Domains       []string            `json:"domains"`
+	// Certificates   map[string]secretRef   `json:"certificates,omitempty"`
```
hugosantos, line 52: Nit: please remove.

schema/server.proto
```diff
@@ -102,8 +104,12 @@ message Server {
         Endpoint.Type            endpoint_type    = 7;
         PackageRef               ingress_provider = 8;
         repeated DomainSpec      ingress_domain   = 10;  // Only has a meaning if this is an ingress.
-
-        reserved 2;
+        repeated IngressAnnotation ingress_annotation = 11;
```
hugosantos, line 107: Would it make sense to use "ServiceAnnotations" here as well?

hugosantos (commented):
Looks good! Just one question.

schema/server.proto
szferi: Yes I was unsure about this but certainly we can

hugosantos (approved):

PR #1097, opened 2023-02-06
provider/nscloud: exchange user tokens to tenant tokens for cluster ops

edganiukov:
Fixes NSL-500

### DIFF
--- a/internal/auth/tokens.go
+++ b/internal/auth/tokens.go
@@ -9,6 +9,7 @@ import (
 	"io/fs"
 	"os"
 	"path/filepath"
+	"strings"
 	"time"
 
 	"github.com/golang-jwt/jwt/v4"
@@ -61,7 +62,7 @@ func LoadTenantToken() (*Token, error) {
 	}
 
 	claims := jwt.RegisteredClaims{}
-	if _, _, err := new(jwt.Parser).ParseUnverified(token.TenantToken, claims); err != nil {
+	if _, _, err := new(jwt.Parser).ParseUnverified(strings.TrimPrefix("nsct_", token.TenantToken), claims); err != nil {
 		return nil, err
 	}
 
--- a/internal/auth/tokens.go
+++ b/internal/auth/tokens.go
@@ -6,9 +6,12 @@ package auth
 
 import (
 	"encoding/json"
+	"io/fs"
 	"os"
 	"path/filepath"
+	"time"
 
+	"github.com/golang-jwt/jwt/v4"
 	"namespacelabs.dev/foundation/internal/fnerrors"
 	"namespacelabs.dev/foundation/internal/workspace/dirs"
 )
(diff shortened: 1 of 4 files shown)


internal/providers/nscloud/api/rpc.go
```diff
@@ -34,22 +34,11 @@ type API struct {
...
-	if err != nil {
-		if os.IsNotExist(err) {
-			// Pass user auth for now.
-			// TODO exchange user auth for tenant tokens.
-			return auth.GenerateToken(ctx)
-		}
-
-		return "", err
+func MakeAPI(endpoint string) API {
+	fetchTenantToken := func(ctx context.Context) (string, error) {
```
n-g, line 38: Let's inline this.

internal/providers/nscloud/api/rpc.go
```diff
@@ -290,6 +285,7 @@ func ExchangeToken(ctx context.Context, scopes ...string) (string, error) {
 		return "", err
 	}
 
+	// No tenant token stored, so use user token and exchange it to a tenant token.
```
n-g, line 288: Should we cache unscoped tenant tokens?

internal/providers/nscloud/api/rpc.go
edganiukov: You mean use `FetchToken: ExchangeToken` the `fnapi.Call` objects? It doesn't work due to type (`ExchangeToken` has optional `scopes` parameter).

n-g: 🙈 yep - true - thanks!

internal/providers/nscloud/api/rpc.go
```diff
@@ -276,8 +265,14 @@ func ListClusters(ctx context.Context, api API) (*KubernetesClusterList, error)
...
 func ExchangeToken(ctx context.Context, scopes ...string) (string, error) {
+	// Check if there is already tenant token stored.
 	tenantToken, err := auth.LoadTenantToken()
 	if err == nil {
+		// If no scopes provided we can immediately return the token.
+		if len(scopes) == 0 {
+			return tenantToken.TenantToken, nil
+		}
+
 		resp, err := fnapi.ExchangeTenantToken(ctx, tenantToken.TenantToken, scopes)
```
hugosantos, line 276: Ideally calls that have non-trivial duration should be wrapped in tasks, so users can see where time is being spent.

internal/providers/nscloud/api/rpc.go
edganiukov: done

internal/auth/tokens.go
```diff
@@ -57,5 +60,18 @@ func LoadTenantToken() (*Token, error) {
 		return nil, err
 	}
 
+	claims := jwt.RegisteredClaims{}
+	if _, _, err := new(jwt.Parser).ParseUnverified(token.TenantToken, claims); err != nil {
+		return nil, err
+	}
+
+	// If stored token is expired, we remove it and return [fs.ErrNotExist].
+	if !claims.VerifyExpiresAt(time.Now(), true) {
```
edganiukov, line 69: Wondering if I should have some leeway here?

nichtverstehen: You mean throw away a token that is still valid, but is about to expire? Would make sense.

But starting simple with a strict check is fine.

(Not the other way of course - we shouldn't try to use recently expired tokens).

edganiukov: > You mean throw away a token that is still valid, but is about to expire? Would make sense.
Yep, that is what I mean.

internal/providers/nscloud/api/rpc.go
edganiukov: done. I hope simply wrapping the whole func body is OK.

n-g (approved):

PR #1108, opened 2023-02-08
auth: remove cached tokens if invalid or corrupted

edganiukov:


n-g (approved):
Shouldn't we preserve the original errors, when clearing the cache succeeds?
I think it would be confusing otherwise that an unmarshal error returns file not found.

edganiukov:
> Shouldn't we preserve the original errors, when clearing the cache succeeds? I think it would be confusing otherwise that an unmarshal error returns file not found.

@n-g So the idea is that if token expired or somehow invalid - we remove the token and return `fs.ErrNotExist` (because its really not exists anymore). `ExchangeToken` function would then check the error and do the exchange if no token cached (as it got removed).

I am now changing the structure of the cache file, which would lead to unmarshaling errors if the file contains old structure, so decided that we should not care about corrupted caches and user should not be aware if this.

n-g:
> > Shouldn't we preserve the original errors, when clearing the cache succeeds? I think it would be confusing otherwise that an unmarshal error returns file not found.
> 
> @n-g So the idea is that if token expired or somehow invalid - we remove the token and return `fs.ErrNotExist` (because its really not exists anymore). `ExchangeToken` function would then check the error and do the exchange if no token cached (as it got removed).
> 
> I am now changing the structure of the cache file, which would lead to unmarshaling errors if the file contains old structure, so decided that we should not care about corrupted caches and user should not be aware if this.

SG, let's wire `ctx context.Context` into here, and mention the original error in the debug output at least :)
E.g. https://github.com/namespacelabs/foundation/blob/1bbb17f9116f6b9c697c4eb436401b0f9f052a47/orchestration/prepare.go#L85-L86

edganiukov:
> SG, let's wire `ctx context.Context` into here, and mention the original error in the debug output at least :) E.g.
> 
> https://github.com/namespacelabs/foundation/blob/1bbb17f9116f6b9c697c4eb436401b0f9f052a47/orchestration/prepare.go#L85-L86

:+1: that is what I also was thinking to do. Thx.

PR #1111, opened 2023-02-09
internal/auth: cache tenant tokens respecting scopes and username

edganiukov:
Fixes NSL-571

### DIFF
--- a/internal/auth/tokens.go
+++ b/internal/auth/tokens.go
@@ -7,10 +7,12 @@ package auth
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io/fs"
 	"os"
 	"path/filepath"
+	"sort"
 	"strings"
 	"time"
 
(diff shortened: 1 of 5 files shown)


internal/auth/tokens.go
```diff
@@ -25,65 +27,98 @@ const (
...
+			return nil, err
+		}
+		return nil, ErrTokenNotExist
 	}
 
 	if !claims.VerifyExpiresAt(time.Now(), true) {
 		fmt.Fprintf(console.Debug(ctx), "tenant JWT has expired\n")
-		return cleanTokenCache(p)
+		if err := os.Remove(f); err != nil {
+			return nil, err
```
n-g, line 97: If a single token is expired, we wipe the entire cache. Shouldn't we just remove/update the expired entry?

internal/auth/tokens.go
```diff
@@ -25,65 +27,98 @@ const (
...
 	if _, _, err := parser.ParseUnverified(strings.TrimPrefix(token.TenantToken, "nsct_"), &claims); err != nil {
 		fmt.Fprintf(console.Debug(ctx), "failed to parse tenant JWT: %v\n", err)
-		return cleanTokenCache(p)
+		if err := os.Remove(f); err != nil {
+			return nil, err
+		}
+		return nil, ErrTokenNotExist
 	}
 
 	if !claims.VerifyExpiresAt(time.Now(), true) {
```
n-g, line 94: Nit: I find this method name very confusing. From what I read, you are doing the right thing, but would be great to make it more obvious.
How about adding a simple wrapper `IsExpired(claims)`?

internal/auth/tokens.go
```diff
@@ -25,65 +27,98 @@ const (
...
+		if err := json.Unmarshal(data, &tokens); err != nil {
+			fmt.Fprintf(console.Debug(ctx), "failed to unmarshal cached tenant tokens: %v\n", err)
+			if err := os.Remove(f); err != nil {
+				return nil, err
+			}
+		}
+
+		return tokens, nil
+	case errors.Is(err, fs.ErrNotExist):
+		return tokens, nil
```
n-g, line 120: nit: for clarity, let's return `cachedTokens{}, nil` here and move `tokens` into the `err == nil` case.

internal/cli/cmd/auth/githubtoken.go
```diff
@@ -49,7 +49,11 @@ func NewExchangeGithubTokenCmd() *cobra.Command {
 				tok = fmt.Sprintf("nsct_%s", res.TenantToken)
 			}
 
-			return auth.StoreTenantToken(tok)
+			if _, err := auth.StoreUser(ctx, &auth.UserAuth{Username: auth.GithubJWTAudience}); err != nil {
```
n-g, line 52: I think we should not store fake users. This will easily confuse us later.
In case of a github workflow, there is no user, but we can still generate tokens.
Can we cache them with an empty user name?

internal/auth/tokens.go
edganiukov: Yeah, true, fixed

internal/cli/cmd/auth/githubtoken.go
edganiukov: Done, doing cache with empty user.

internal/auth/tokens.go
edganiukov: Done

internal/auth/tokens.go
edganiukov: Added

internal/auth/tokens.go
```diff
@@ -25,65 +27,98 @@ const (
...
+	cache, err := getCachedTokens(ctx, f)
 	if err != nil {
 		return err
 	}
 
-	if err := os.WriteFile(filepath.Join(configDir, tokenTxt), data, 0600); err != nil {
-		return fnerrors.New("failed to write token data: %w", err)
+	sort.Strings(token.Scopes)
+	cache[username+strings.Join(token.Scopes, "")] = token
+
```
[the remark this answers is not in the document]
edganiukov: Changed to cache as a list of tokens with metadata.

internal/auth/tokens.go
```diff
@@ -25,65 +27,147 @@ const (
...
+	Username    string   `json:"username,omitempty"`
+	Scopes      []string `json:"scopes,omitempty"`
+	TenantToken string   `json:"tenant_token,omitempty"`
 }
 
-func StoreTenantToken(token string) error {
-	data, err := json.Marshal(Token{TenantToken: token})
-	if err != nil {
-		return err
+func (t Token) Equal(other Token) bool {
```
n-g, line 38: nit: s/Equal/EqualScope

internal/auth/tokens.go
```diff
@@ -25,65 +27,147 @@ const (
...
+	return Token{}, false
+}
+
+func (c *cachedTokens) deleteToken(token Token) {
+	idx := slices.IndexFunc(c.tokens, token.Equal)
+	if idx == -1 {
+		return
+	}
+
+	temp := c.tokens[:idx]
```
n-g, line 136: https://pkg.go.dev/golang.org/x/exp/slices#Delete ?

internal/auth/tokens.go
hugosantos, line 38: Nit: Equals for consistency with other APIs.

internal/auth/tokens.go
```diff
@@ -25,65 +27,147 @@ const (
...
+	TenantToken string   `json:"tenant_token,omitempty"`
 }
 
-func StoreTenantToken(token string) error {
-	data, err := json.Marshal(Token{TenantToken: token})
-	if err != nil {
-		return err
+func (t Token) Equal(other Token) bool {
+	if t.Username == other.Username {
+		slices.Sort(t.Scopes)
```
hugosantos, line 40: A method like `Equals` should not mutate its state.

Either pre-sort, or clone the slice to sort.

internal/providers/nscloud/api/rpc.go
```diff
@@ -292,45 +293,58 @@ func ListClusters(ctx context.Context, api API) (*KubernetesClusterList, error)
...
-		tenantToken, err := auth.LoadTenantToken(ctx)
+		var username string
+		userAuth, err := auth.LoadUser()
 		if err == nil {
-			// If no scopes provided we can immediately return the token.
-			if len(scopes) == 0 {
-				return tenantToken.TenantToken, nil
-			}
+			username = userAuth.Username
+		} else if !(errors.Is(err, auth.ErrRelogin) && os.Getenv("GITHUB_ACTIONS") == "true") {
```
n-g, line 300: Nit: `os.Getenv("GITHUB_ACTIONS") == "true"` should be a function in `internal/github`, similar to https://github.com/namespacelabs/foundation/blob/63ec8aa642dd072d658d8532a090cc6c08b2abcd/internal/environment/ci.go#L10-L11

internal/providers/nscloud/api/rpc.go
```diff
@@ -292,45 +293,58 @@ func ListClusters(ctx context.Context, api API) (*KubernetesClusterList, error)
...
+
+		scopedToken, err := auth.LoadTenantToken(ctx, username, scopes)
+		if err == nil {
+			// The tenant token with requested scopes is in cache.
+			return scopedToken.TenantToken, nil
+		}
 
+		var token string
+		// If no scoped token cached, then first try exchange unscoped tenant token and otherwise use user token.
+		tenantToken, err := auth.LoadTenantToken(ctx, username, nil)
```
hugosantos, line 313: Username is a past concept, lets not carry it over. :-)

n-g (approved):
Thanks for the edits!

internal/providers/nscloud/api/rpc.go
edganiukov: We discussed the use-case when user logins as a different user, I still need some identity to attach tenant tokens to.

hugosantos:
@edganiukov lets have a quick chat about this.

PR #1114, opened 2023-02-10
internal/auth: cache tenant token on login

edganiukov:
Also in this change:
- In admin mode we exchange user token to a tenant token with an admin scope.
- Changed registry to use the cached tenant token without exchanged to a scoped token.

### DIFF
--- a/internal/cli/cmd/auth/login.go
+++ b/internal/cli/cmd/auth/login.go
@@ -60,7 +60,7 @@ func NewLoginCmd() *cobra.Command {
 				return err
 			}
 
-			tt, err := fnapi.ExchangeUserToken(ctx, userToken, nil)
+			tt, err := fnapi.ExchangeUserToken(ctx, userToken)
 			if err != nil {
 				return err
 			}
--- a/internal/cli/cmd/auth/login.go
+++ b/internal/cli/cmd/auth/login.go
@@ -55,6 +55,20 @@ func NewLoginCmd() *cobra.Command {
 				return err
 			}
 
+			userToken, err := auth.GenerateTokenFromUserAuth(ctx, userAuth)
+			if err != nil {
+				return err
+			}
+
+			tt, err := fnapi.ExchangeUserToken(ctx, userToken, nil)
+			if err != nil {
+				return err
+			}
+
+			if err := auth.StoreTenantToken(tt.TenantToken); err != nil {
+				return err
+			}
+
 			fmt.Fprintf(stdout, "\nHi %s, you are now logged in, have a nice day.\n", username)
 
 			return nil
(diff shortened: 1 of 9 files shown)


internal/providers/nscloud/api/rpc.go
```diff
@@ -102,6 +101,31 @@ func MakeAPI(endpoint string) API {
 	}
 }
 
+func FetchTenantToken(ctx context.Context) (*auth.Token, error) {
+	return tasks.Return(ctx, tasks.Action("nscloud.fetch-tenant-token"), func(ctx context.Context) (*auth.Token, error) {
+		if !fnapi.AdminMode {
+			t, err := auth.LoadTenantToken(ctx)
+			if err != nil {
+				return nil, err
+			}
```
n-g, line 110: `return auth.LoadTenantToken(ctx)`

internal/providers/nscloud/api/rpc.go
```diff
@@ -102,6 +101,31 @@ func MakeAPI(endpoint string) API {
...
+			return t, nil
+		}
+
+		// In admin mode we exchange user token to a tenant token with `admin` scope.
+		userToken, err := auth.GenerateToken(ctx)
+		if err != nil {
+			return nil, err
+		}
+
+		t, err := fnapi.ExchangeUserToken(ctx, userToken, []string{"admin"})
```
n-g, line 120: Let's change `ExchangeUserToken` to receive scopes as `...string` and add a string constant `AdminScope`

internal/cli/cmd/auth/login.go
```diff
@@ -55,6 +55,20 @@ func NewLoginCmd() *cobra.Command {
 				return err
 			}
 
+			userToken, err := auth.GenerateTokenFromUserAuth(ctx, userAuth)
```
n-g, line 58: If this fails, we can have an old token cached and a new user.
Let's wipe the cached user + token at the beginning of `ns login`

internal/providers/nscloud/api/rpc.go
edganiukov: Done.

internal/providers/nscloud/api/rpc.go
edganiukov: True, looks better. Done.

internal/cli/cmd/auth/login.go
edganiukov: Do you prefer to remove files or store empty objects?

n-g: Deleting the files feels very clean to me.

edganiukov: Done

n-g (approved):

PR #1121, opened 2023-02-16
cli/nsc: tail and get cluster logs

edganiukov:


internal/cli/cmd/cluster/cmd.go
```diff
@@ -356,6 +357,72 @@ func newHistoryCmd() *cobra.Command {
...
+		Short: "Prints log for a cluster.",
+		Args:  cobra.MaximumNArgs(1),
+	}
+
+	follow := cmd.Flags().BoolP("follow", "f", false, "Specify if the logs should be streamed.")
+	namespace := cmd.Flags().StringP("namespace", "n", "", "Print the logs of this namespace.")
+	pod := cmd.Flags().StringP("pod", "p", "", "Print the logs of this pod.")
+	container := cmd.Flags().StringP("container", "c", "", "Print the logs of this container.")
+	since := cmd.Flags().Duration("since", time.Duration(0), "Show logs since a relative timestamp (e.g. 42m for 42 minutes). Not used for tailing logs.")
+	until := cmd.Flags().Duration("until", time.Duration(0), "Show logs until a relative timestamp (e.g. 42m for 42 minutes). Not used for tailing logs.")
```
n-g, line 372: Let's err out if these are set together with `-f`

internal/providers/nscloud/api/types.go
```diff
@@ -81,6 +87,46 @@ type KubernetesCluster struct {
...
+	Line      []LogLine `json:"line,omitempty"`
+}
+
+type LogLine struct {
+	Timestamp time.Time `json:"timestamp,omitempty"`
+	Content   string    `json:"content,omitempty"`
+	Stream    string    `json:"stream,omitempty"`
+}
+
+func (l LogLine) Fprint(w io.Writer) {
```
n-g: I'd find it more natural to offer `.String()`

internal/cli/cmd/cluster/cmd.go
n-g, line 372: Maybe we should only offer `since` for now, and always pin the end to now.

internal/providers/nscloud/api/rpc.go
```diff
@@ -340,6 +355,80 @@ func ListClusters(ctx context.Context, api API, previousRuns bool) (*KubernetesC
...
+			return err
+		}
+		return nil
+	})
+}
+
+func GetClusterLogs(ctx context.Context, api API, opts *LogsOpts) (*GetLogsResponse, error) {
+	return tasks.Return(ctx, tasks.Action("nscloud.get-cluster-logs"), func(ctx context.Context) (*GetLogsResponse, error) {
+		var selector *LogsSelector
+		if opts.Namespace != "" || opts.Pod != "" || opts.Container != "" {
```
n-g, line 404: You can set them unconditionally, the API skips empty strings.

n-g (approved):

internal/cli/cmd/cluster/cmd.go
edganiukov: done

internal/providers/nscloud/api/types.go
edganiukov: done

PR #1126, opened 2023-02-21
Use tenant token in naming API

gmichelo:
Fixes: NSL-632


internal/fnapi/tenants.go
```diff
@@ -67,3 +72,24 @@ func ExchangeTenantToken(ctx context.Context, token string, scopes []string) (Ex
 
 	return res, nil
 }
+
+func FetchTenantToken(ctx context.Context) (*auth.Token, error) {
+	return tasks.Return(ctx, tasks.Action("nscloud.fetch-tenant-token"), func(ctx context.Context) (*auth.Token, error) {
```
n-g, line 77: "nscloud.fetch-tenant-token" -> "api.fetch-tenant-token"?

internal/fnapi/naming.go
```diff
@@ -83,14 +83,22 @@ func AllocateName(ctx context.Context, opts AllocateOpts) (*NameResource, error)
...
+
 		if err := (Call[IssueRequest]{
 			Endpoint: EndpointAddress,
 			Method:   "nsl.naming.NamingService/Issue",
 			PreAuthenticateRequest: func(ua *auth.UserAuth, rt *IssueRequest) error {
 				rt.UserAuth = ua
 				return nil
 			},
-			FetchToken: auth.GenerateToken,
+			FetchToken: fetchTenantToken,
```
n-g, line 101: We also need to update https://github.com/namespacelabs/foundation/blob/0ba4493266b26d72c9a9f2959824e347189139bd/internal/fnapi/map.go#L35-L36

internal/fnapi/naming.go
```diff
@@ -83,14 +83,22 @@ func AllocateName(ctx context.Context, opts AllocateOpts) (*NameResource, error)
 
 		var nr IssueResponse
 
+		fetchTenantToken := func(ctx context.Context) (string, error) {
```
n-g, line 86: After moving Naming to tenant tokens, the only remaining calls using user auth tokens are Telemetry and Versions.
Both of these set `OptionalAuth = true` and indeed do not need the token.
So I'd suggest removing the FetchToken from them and updating the type at 
https://github.com/namespacelabs/foundation/blob/0ba4493266b26d72c9a9f2959824e347189139bd/internal/fnapi/fnapi.go#L63-L64
To be `func(ctx context.Context) (*auth.Token, error)`

This would allow us to skip the temp function here and in `rpc.go`

internal/fnapi/tenants.go
gmichelo: I opted for "tenants.fetch-tenant-token", as other actions in the same package don't use `api.*` prefix, but rather something more specific (e.g. `dns.*` for naming actions)

internal/fnapi/fnapi.go
```diff
@@ -60,7 +60,7 @@ type Call[RequestT any] struct {
 	Method                 string
 	PreAuthenticateRequest func(*auth.UserAuth, *RequestT) error
 	OptionalAuth           bool // Don't fail if not authenticated.
```
n-g, line 62: Now that telemetry and versions won't attach tokens, we can remove `OptionalAuth`

n-g (approved):
Looks good!
Please only merge after manually testing your change against an already updated prod API server :)

PR #1134, opened 2023-02-28
nsc: attach cluster creator from CLI

n-g:
Fixes NSL-658


schema/nscloud/attachments.proto
```diff
@@ -0,0 +1,13 @@
...
+// you may not use this file except in compliance with the License.
+
+syntax = "proto3";
+
+package foundation.schema.nscloud;
+
+option go_package = "namespacelabs.dev/foundation/schema/nscloud";
+
+message CreatorAttachment {
+    string username = 1;
```
nichtverstehen, line 12: I don't think we ever defined what an ns username means. Should we have the user ID here instead?

PR #1140, opened 2023-03-02
Log actions to a file and capture the artifact.

n-g:
Fixes NSL-678

### DIFF
--- a/std/tasks/simplelog/logger.go
+++ b/std/tasks/simplelog/logger.go
@@ -8,7 +8,6 @@ import (
 	"bytes"
 	"fmt"
 	"io"
-	"os"
 
 	"github.com/spf13/pflag"
 	"namespacelabs.dev/foundation/internal/console/colors"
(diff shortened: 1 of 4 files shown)


std/tasks/simplelog/logger.go
```diff
@@ -38,6 +53,24 @@ func (sl *logger) Waiting(ra *tasks.RunningAction) {
...
 }
 
+func (sl logger) write(b []byte) {
+	if LogActionsToFile == "" {
+		// Ignore errors
+		_, _ = sl.out.Write(b)
+		return
+	}
+
+	fo, err := os.OpenFile(LogActionsToFile, os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0644)
```
hugosantos, line 63: What about opening the file on NewSink, and set sl.out?

n-g: I tried this first. My impression is that there is an ordering problem (we call NewSink before evaluating flags).

gmichelo (approved):

PR #1142, opened 2023-03-04
rename experimentalFromSecret to fromKubernetesSecret to better reflect its functionality and make it official.

szferi:


internal/frontend/cuefrontend/args/resolvable.go
```diff
@@ -38,7 +38,7 @@ var validKeys = []string{
 	"fromServiceEndpoint",
 	"fromServiceIngress",
 	"fromResourceField",
-	"experimentalFromSecret",
+	"fromKubernetesSecret",
```
hugosantos, line 41: It's good that internally we switch to fromKubernetesSecret, but lets keep `experimentalFromSecret` around in the frontend for a bit, as there are uses of it that would be broken by this change. You can add a 2nd case here and have both parse into ev.fromKubernetesSecret.

hugosantos (commented):
LG but have a small request!

internal/frontend/cuefrontend/args/resolvable.go
szferi: Sure, I updated the PR

PR #1156, opened 2023-03-15
internal/auth: Avoid resetting login state before new login succeeds.

nichtverstehen:
I often run into the inconvenience when I run `ns login` by mistake. And it immediately resets the login state, so I cannot just cancel the login process and keep the old auth.

### DIFF
--- a/internal/auth/tokens.go
+++ b/internal/auth/tokens.go
@@ -89,16 +89,3 @@ func LoadTenantToken(ctx context.Context) (*Token, error) {
 
 	return token, nil
 }
-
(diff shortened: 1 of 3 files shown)


edganiukov (approved):
To me its fine, but we had a discussion about it here https://github.com/namespacelabs/foundation/pull/1114#discussion_r1102736554

nichtverstehen:
Thanks for the context! The original concern was partial updates (tenant and user tokens go out of sync). It's not a danger in the new login flow. So I'll merge this.

PR #1163, opened 2023-03-20
Ensure the cmd name is evaluate at error generation time.

n-g:
Fixes NSL-807


internal/cli/cmd/auth/githubtoken.go
```diff
@@ -34,10 +35,11 @@ func NewExchangeGithubTokenCmd() *cobra.Command {
 		}
 
 		if *ensuredDuration > 0 {
+			var relogin *fnerrors.ReloginErr
 			if err := auth.EnsureTokenValidAt(ctx, time.Now().Add(*ensuredDuration)); err == nil {
 				// Token is valid for entire duration.
 				return nil
-			} else if err != auth.ErrRelogin {
+			} else if errors.As(err, &relogin) {
```
edganiukov, line 42: As it was `if err != auth.ErrRelogin`, here you need change to `if !errors.As(err, &relogin)`

edganiukov (changes requested):

gmichelo (approved):

PR #1167, opened 2023-03-27
cmd: proxy: create arm64 cluster

gmichelo:


internal/providers/nscloud/api/rpc.go
```diff
@@ -221,10 +221,12 @@ func CreateAndWaitCluster(ctx context.Context, api API, opts CreateClusterOpts)
 	return WaitCluster(ctx, api, cluster.ClusterId, opts.WaitClusterOpts)
 }
 
-func EnsureBuildCluster(ctx context.Context, api API) (*CreateClusterResult, error) {
+func EnsureBuildCluster(ctx context.Context, api API, features ...string) (*CreateClusterResult, error) {
```
edganiukov, line 224: nit: would be nice to create a struct `EnsureBuildClusterOpts` (to follow the pattern), but non-blocking.

edganiukov (approved):

PR #1178, opened 2023-04-26
nsc/cred-helper: docker-login installs docker-credential-nsc helper

gmichelo:
Tested:
```
{
	"auths": {
		"ghcr.io": {},
		"https://index.docker.io/v1/": {}
	},
	"credsStore": "desktop",
	"credHelpers": {
    		"nscr.io": "nsc"
  	}
}
```

```
docker pull nscr.io/<workspace-id>/ns-go-app:v0.0.1
```

### DIFF
--- a/internal/cli/cmd/cluster/docker.go
+++ b/internal/cli/cmd/cluster/docker.go
@@ -99,8 +99,8 @@ func NewDockerLoginCmd(hidden bool) *cobra.Command {
 
 		cfg := config.LoadDefaultConfigFile(console.Stderr(ctx))
 
-		for _, x := range []*api.ImageRegistry{response.Registry, response.NSCR} {
-			if x != nil {
+		for _, reg := range []*api.ImageRegistry{response.Registry, response.NSCR} {
+			if reg != nil {
 				if err := cfg.GetCredentialsStore(response.Registry.EndpointAddress).Store(types.AuthConfig{
 					ServerAddress: x.EndpointAddress,
 					Username:      dockerUsername,
@@ -186,7 +186,7 @@ func NewDockerCredHelperGetCmd(hidden bool) *cobra.Command {
 		}
 
 		for _, reg := range []*api.ImageRegistry{resp.Registry, resp.NSCR} {
-			if regURL == reg.EndpointAddress {
+			if reg != nil && regURL == reg.EndpointAddress {
 				token, err := fnapi.FetchTenantToken(ctx)
 				if err != nil {
 					return fnerrors.New("failed to fetch tenant token: %w", err)
@@ -235,7 +235,9 @@ func NewDockerCredHelperListCmd(hidden bool) *cobra.Command {
 		output := map[string]string{}
 
 		for _, reg := range []*api.ImageRegistry{resp.Registry, resp.NSCR} {
-			output[reg.EndpointAddress] = dockerUsername
+			if reg != nil {
+				output[reg.EndpointAddress] = dockerUsername
+			}
 		}
 
 		buf, err := json.Marshal(output)
--- a/internal/cli/cmd/cluster/docker.go
+++ b/internal/cli/cmd/cluster/docker.go
@@ -30,7 +30,8 @@ import (
 
 const (
 	dockerUsername   = "tenant-token"
-	credHelperBinary = "nsc"
+	credHelperBinary = "docker-credentials-nsc"
+	nscBinary        = "nsc"
 )
 
 func newDockerCmd() *cobra.Command {
@@ -102,7 +103,7 @@ func NewDockerLoginCmd(hidden bool) *cobra.Command {
 
 		for _, reg := range []*api.ImageRegistry{response.Registry, response.NSCR} {
 			if reg != nil {
-				cfg.CredentialHelpers[reg.EndpointAddress] = credHelperBinary
+				cfg.CredentialHelpers[reg.EndpointAddress] = nscBinary
 			}
 		}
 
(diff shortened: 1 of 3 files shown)


internal/cli/cmd/cluster/docker.go
```diff
@@ -126,15 +129,158 @@ func NewDockerLoginCmd(hidden bool) *cobra.Command {
 			return fnerrors.New("failed to write %q: %w", cfgFile, err)
 		}
 
-		if nscr := response.NSCR; nscr != nil {
-			fmt.Fprintf(stdout, "\nYou are now logged into your Workspace container registry:\n\n  %s/%s", nscr.EndpointAddress, nscr.Repository)
-			fmt.Fprintf(stdout, "\n\nRun your first build with:\n\n  $ nsc build . -n test:v0.0.1 -p")
+		if _, err := exec.LookPath(credHelperBinary); err != nil {
+			if errors.Is(err, exec.ErrNotFound) {
+				fmt.Fprintf(stdout, "\nWarning: nsc binary is not installed in your $PATH. Docker credentials won't work until you add nsc's path to the $PATH.\n")
```
hugosantos, line 134: If we modify the configuration, then I suggest always printing the block below, with next steps.

And adding a warning at the end (as that's last thing that people read). And I would suggest something more like:

"We didn't find docker-credentials-nsc in your $PATH. It's usually installed along-side nsc; so if you have nsc to the $PATH, docker-credentials-nsc will also be available.

While your $PATH is not updated, accessing nscr.io images from docker-based tools won't work. But you can always use `nsc build` (as per above) or `nsc run`."

internal/cli/cmd/cluster/docker.go
```diff
@@ -126,15 +129,158 @@ func NewDockerLoginCmd(hidden bool) *cobra.Command {
...
+		return fnerrors.New("unimplemented")
+	})
+
+	return cmd
+}
+
+func NewDockerCredHelperGetCmd(hidden bool) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:    "get",
+		Short:  "Get Workspace's container registry credetial",
```
hugosantos, line 171: credentials

internal/cli/cmd/cluster/docker.go
```diff
@@ -126,15 +129,158 @@ func NewDockerLoginCmd(hidden bool) *cobra.Command {
...
+					return fnerrors.New("failed to marshal output JSON: %w", err)
+				}
+
+				fmt.Println(string(buf))
+				return nil
+			}
+		}
+
+		return fnerrors.New("credentials not found in nscloud")
+
```
hugosantos, line 215: remove extra line

internal/cli/cmd/cluster/docker.go
```diff
@@ -126,15 +129,158 @@ func NewDockerLoginCmd(hidden bool) *cobra.Command {
...
+				if err != nil {
+					return fnerrors.New("failed to marshal output JSON: %w", err)
+				}
+
+				fmt.Println(string(buf))
+				return nil
+			}
+		}
+
+		return fnerrors.New("credentials not found in nscloud")
```
hugosantos, line 214: Looking at the docker implementation, it seems that docker surfaces the command output to the user, when there are failures. Which is good.

It also has special-casing for some messages. For examples, if we don't handle the specified server, we should output "credentials not found in native keychain".

internal/cli/cmd/cluster/docker.go
```diff
@@ -126,15 +129,158 @@ func NewDockerLoginCmd(hidden bool) *cobra.Command {
...
+		resp, err := api.GetImageRegistry(ctx, api.Endpoint)
+		if err != nil {
+			return fnerrors.New("failed to get nscloud registries: %w", err)
+		}
+
+		for _, reg := range []*api.ImageRegistry{resp.Registry, resp.NSCR} {
+			if reg != nil && regURL == reg.EndpointAddress {
+				token, err := fnapi.FetchTenantToken(ctx)
+				if err != nil {
+					return fnerrors.New("failed to fetch tenant token: %w", err)
```
hugosantos, line 195: I believe this may be a "please login" message, so I suggest not wrapping it.

internal/cli/cmd/cluster/docker.go
```diff
@@ -126,15 +129,158 @@ func NewDockerLoginCmd(hidden bool) *cobra.Command {
...
+					Username:  dockerUsername,
+					Secret:    token.Raw(),
+				}
+
+				buf, err := json.Marshal(c)
+				if err != nil {
+					return fnerrors.New("failed to marshal output JSON: %w", err)
+				}
+
+				fmt.Println(string(buf))
```
hugosantos, line 209: Use, instead:

enc := json.NewEncoder(os.Stdout)
return enc.Encode(c)

internal/cli/cmd/cluster/docker.go
```diff
@@ -126,15 +129,158 @@ func NewDockerLoginCmd(hidden bool) *cobra.Command {
...
+
+	})
+
+	return cmd
+}
+
+func NewDockerCredHelperListCmd(hidden bool) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:    "list",
+		Short:  "List Workspace's container registry credetials",
```
hugosantos, line 224: credentials

internal/cli/cmd/cluster/docker.go
```diff
@@ -126,15 +129,158 @@ func NewDockerLoginCmd(hidden bool) *cobra.Command {
...
+				output[reg.EndpointAddress] = dockerUsername
+			}
+		}
+
+		buf, err := json.Marshal(output)
+		if err != nil {
+			return fnerrors.New("failed to marshal output JSON: %w", err)
+		}
+
+		fmt.Println(string(buf))
```
hugosantos, line 251: Same as above: re outputting json.

internal/cli/cmd/cluster/docker.go
```diff
@@ -126,15 +129,158 @@ func NewDockerLoginCmd(hidden bool) *cobra.Command {
...
+func NewDockerCredHelperEraseCmd(hidden bool) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:    "erase",
+		Short:  "Unimplemented",
+		Args:   cobra.NoArgs,
+		Hidden: hidden,
+	}
+
+	cmd.RunE = fncobra.RunE(func(ctx context.Context, args []string) error {
+		return fnerrors.New("unimplemented")
```
hugosantos, line 267: Lets say "not supported" instead; "unimplemented" is usually a todo.

hugosantos (approved):

PR #1182, opened 2023-05-31
cluster/buildx: cleanup background proxy with state management

gmichelo:
* State file to stable location
* Add node group name to metadata.json, so cleanup command can read it and avoid asking user for `--name` input
* Add background process ID to metadata.json, so cleanup command can read it and seng SIGTERM signal
* Implement SIGTERM handler to buildx proxy
* Print informative message even when buildx proxy is run in background
* Delete state folder when cleanup command is called
* Print error message if state metadata.json file is found

Fixes: NSL-1137

### DIFF
--- a/internal/cli/cmd/cluster/buildx.go
+++ b/internal/cli/cmd/cluster/buildx.go
@@ -8,10 +8,8 @@ import (
 	"context"
 	"fmt"
 	"os"
-	"os/signal"
 	"path/filepath"
 	"strings"
-	"syscall"
 
 	"github.com/docker/buildx/store"
 	"github.com/docker/buildx/store/storeutil"
@@ -63,25 +61,7 @@ func newSetupBuildxCmd(cmdName string) *cobra.Command {
 			return err
 		}
 
-		ctx, cancel := context.WithCancel(ctx)
-		defer cancel()
-
 		eg := executor.New(ctx, "proxies")
-
-		eg.Go(func(ctx context.Context) error {
-			// Listen to SIGTERM signal
-			sigCh := make(chan os.Signal, 1)
-			signal.Notify(sigCh, syscall.SIGTERM)
-
-			select {
-			case <-ctx.Done():
-			case <-sigCh:
-				// Graceful termination
-				cancel()
-			}
-
-			return nil
-		})
 
 		available, err := determineAvailable(ctx)
 		if err != nil {
@@ -109,9 +89,8 @@ func newSetupBuildxCmd(cmdName string) *cobra.Command {
 		for _, p := range available {
 			sockPath := filepath.Join(state, fmt.Sprintf("%s.sock", p))
 			md.Instances = append(md.Instances, buildxInstanceMetadata{
-				Platform:           p,
-				SocketPath:         sockPath,
-				BackgroundProxyPID: -1, // -1 means proxy is not in background
+				Platform:   p,
+				SocketPath: sockPath,
 			})
 		}
 
@@ -125,9 +104,10 @@ func newSetupBuildxCmd(cmdName string) *cobra.Command {
 				if pid, err := startBackgroundProxy(ctx, p, *createAtStartup); err != nil {
 					return err
 				} else {
-					md.Instances[i].BackgroundProxyPID = pid
+					md.Instances[i].Pid = pid
 				}
 			} else {
+				md.Instances[i].Pid = os.Getpid()
 				bp, err := instance.runBuildProxy(ctx, p.SocketPath)
 				if err != nil {
 					return err
@@ -187,6 +167,10 @@ func newSetupBuildxCmd(cmdName string) *cobra.Command {
 			return err
 		}
 
+		if err := os.RemoveAll(state); err != nil {
+			return err
+		}
+
 		return nil
 	})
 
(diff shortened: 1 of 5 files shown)


internal/cli/cmd/cluster/buildx.go
```diff
@@ -56,8 +63,26 @@ func newSetupBuildxCmd(cmdName string) *cobra.Command {
 			return err
 		}
 
+		ctx, cancel := context.WithCancel(ctx)
```
hugosantos, line 66: See also https://github.com/namespacelabs/foundation/blob/fe39ff7287121c5ef3da97c94055953046684c4b/internal/cli/fncobra/sigint.go#L16-L17 for an example of a composable version of this.

internal/cli/cmd/cluster/buildx.go
```diff
@@ -67,34 +92,40 @@ func newSetupBuildxCmd(cmdName string) *cobra.Command {
...
+			NodeGroupName: *name,
+		}
 		for _, p := range available {
 			sockPath := filepath.Join(state, fmt.Sprintf("%s.sock", p))
 			md.Instances = append(md.Instances, buildxInstanceMetadata{
-				Platform:   p,
-				SocketPath: sockPath,
+				Platform:           p,
+				SocketPath:         sockPath,
+				BackgroundProxyPID: -1, // -1 means proxy is not in background
```
hugosantos, line 114: pid 0 is also not defined so I'd lean on using that as the marker of pid not being set.

internal/cli/cmd/cluster/buildx.go
```diff
@@ -157,26 +193,49 @@ func newSetupBuildxCmd(cmdName string) *cobra.Command {
...
+	NodeGroupName string                   `json:"node_group_name"`
+	Instances     []buildxInstanceMetadata `json:"instances"`
 }
 
 type buildxInstanceMetadata struct {
-	Platform   api.BuildPlatform `json:"build_platform"`
-	SocketPath string            `json:"socket_path"`
+	Platform           api.BuildPlatform `json:"build_platform"`
+	SocketPath         string            `json:"socket_path"`
+	BackgroundProxyPID int               `json:"bg_proxy_pid"`
```
hugosantos, line 238: Nit: I think naming this simply "pid" would be fine.

internal/cli/cmd/cluster/buildx.go
```diff
@@ -246,7 +305,40 @@ func newCleanupBuildxCommand() *cobra.Command {
...
+
+			for _, inst := range md.Instances {
+				if inst.BackgroundProxyPID > 0 {
+					process, err := os.FindProcess(inst.BackgroundProxyPID)
+					if err != nil {
+						return err
+					}
+
+					err = process.Signal(syscall.SIGTERM)
+					if err != nil && !errors.Is(err, os.ErrProcessDone) {
```
hugosantos, line 331: Avoid re-assignments:

`if err := process.Signal(syscall.SIGTERM); err != nil && !errors.Is(err, ...) {`

internal/cli/cmd/cluster/buildx.go
```diff
@@ -246,7 +305,40 @@ func newCleanupBuildxCommand() *cobra.Command {
...
+						return err
+					}
+				}
+			}
+
+			if err := os.RemoveAll(state); err != nil {
+				return err
```
hugosantos, line 338: I'd make this a warning, to not skip txn.Remove as well.

internal/cli/cmd/cluster/buildx.go
```diff
@@ -246,7 +305,40 @@ func newCleanupBuildxCommand() *cobra.Command {
...
+			}
+
+			return txn.Remove(md.NodeGroupName)
```
hugosantos, line 341: `if md.NodeGroupName != "" {`?

internal/cli/cmd/cluster/buildx.go
```diff
@@ -318,11 +410,19 @@ func banner(ctx context.Context, name string, use bool, native []api.BuildPlatfo
...
 
-	fmt.Fprintln(w)
-	fmt.Fprintln(w, style.Comment.Apply("Exiting will remove the configuration."))
+		fmt.Fprintln(w)
+		fmt.Fprintln(w, style.Comment.Apply("Exiting will remove the configuration."))
+	} else {
+		fmt.Fprintf(w, "\nStart building:\n")
+		fmt.Fprintf(w, "\n  docker buildx build ...\n")
+
+		fmt.Fprintf(w, "\nThe proxy is running in background. You can stop it and unconfigure the remote builders with:\n")
```
hugosantos, line 423: "proxy" is an implementation detail, I'd avoid making references to it. I suggest saying:

"Your remote builder context is running in the background. You can always clean up previous runs with: nsc docker buildx cleanup".

hugosantos (approved):

internal/cli/cmd/cluster/buildx.go
gmichelo: Ah! Thanks for the reference. I see that `WithSigIntCancel` is already wired in 
https://github.com/namespacelabs/foundation/blob/cleanup-buildx-proxy/internal/cli/fncobra/do.go#L67
Which eventually is used by 
https://github.com/namespacelabs/foundation/blob/cleanup-buildx-proxy/internal/cli/fncobra/main.go#L62

So we can remove this bit and change the signal from SIGTERM to SIGINT.

PR #1190, opened 2023-06-06
legal: require signing of CLA by external contributors

ghuntley:
Opening a proposal that external contributors are required to sign a CLA (contents undefined) which is enforced via GitHub Actions.

Next steps:
- Discuss implementation of a CLA
- Create https://github.com/namespacelabs/cla repo
- Create CLA document at https://github.com/namespacelabs/cla/README.md
- Create `CLA_GITHUB_TOKEN` token which has write access to `namespacelabs/cla`

See https://github.com/coder/cla/tree/main as an example of this running in the wild

Each time `https://github.com/namespacelabs/cla/README.md` is updated then the workflow should bump the vYYYY-MM-DD` field in the GitHub Actions yaml to ensure folks read/review/sign with the updated terms.

### DIFF
--- a/.github/workflows/cla.yaml
+++ b/.github/workflows/cla.yaml
@@ -1,0 +1,26 @@
+name: "CLA Assistant"
+on:
+  issue_comment:
+    types: [created]
+  pull_request_target:
+    types: [opened,closed,synchronize]
+
+jobs:
+  CLAssistant:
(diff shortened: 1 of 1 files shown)


hugosantos:
This makes sense! Want to get back to it over the next days.

PR #1206, opened 2023-08-28
Buildkit grpc proxy

gmichelo:
Fixes: NSL-1536

### DIFF
--- a/internal/cli/cmd/cluster/buildproxy.go
+++ b/internal/cli/cmd/cluster/buildproxy.go
@@ -40,9 +40,9 @@ type BuildClusterInstance struct {
 	cancelRefresh func()
 }
 
-func (bp *BuildClusterInstance) NewConn(ctx context.Context) (net.Conn, error) {
+func (bp *BuildClusterInstance) NewConn(parentCtx context.Context) (net.Conn, error) {
 	// Wait at most 3 minutes to create a connection to a build cluster.
-	ctx, done := context.WithTimeout(ctx, 3*time.Minute)
+	ctx, done := context.WithTimeout(parentCtx, 3*time.Minute)
 	defer done()
 
 	// This is not our usual play; we're doing a lot of work with the lock held.
@@ -67,12 +67,12 @@ func (bp *BuildClusterInstance) NewConn(ctx context.Context) (net.Conn, error) {
 
 	if bp.previous == nil || bp.previous.ClusterId != response.ClusterId {
 		if err := waitUntilReady(ctx, response); err != nil {
-			return nil, fmt.Errorf("Failed to wait for buildkit to become ready: %w\n", err)
+			return nil, fmt.Errorf("failed to wait for buildkit to become ready: %w", err)
 		}
 	}
 
 	if response.BuildCluster != nil && !response.BuildCluster.DoesNotRequireRefresh {
-		bp.cancelRefresh = api.StartBackgroundRefreshing(ctx, response.ClusterId)
+		bp.cancelRefresh = api.StartBackgroundRefreshing(parentCtx, response.ClusterId)
 	}
 
 	bp.previous = response
--- a/internal/cli/cmd/cluster/buildproxy.go
+++ b/internal/cli/cmd/cluster/buildproxy.go
@@ -25,6 +25,7 @@ import (
 	"namespacelabs.dev/foundation/internal/fnapi"
 	"namespacelabs.dev/foundation/internal/providers/nscloud/api"
 	"namespacelabs.dev/foundation/internal/workspace/dirs"
+	"namespacelabs.dev/foundation/std/tasks"
 )
 
 const (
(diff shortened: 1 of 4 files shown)


internal/cli/cmd/cluster/grpcproxy.go
```diff
@@ -0,0 +1,199 @@
...
+	if err != nil {
+		fmt.Fprintf(console.Errors(context.Background()), "failed to create client stream: %v\n", err)
+		return status.Errorf(codes.Internal, "failed create client: %v", err)
+	}
+
+	s2cErrChan := g.proxyServerToClient(serverStream, clientStream)
+	c2sErrChan := g.proxyClientToServer(clientStream, serverStream)
+	// Make sure to close both client and server connections
+	for i := 0; i < 2; i++ {
+		select {
```
n-g, line 123: Should there be a case `<-serverStream.Context().Done()` for a clean cancellation?

internal/cli/cmd/cluster/grpcproxy.go
```diff
@@ -0,0 +1,199 @@
...
+		}
+	}
+
+	return status.Errorf(codes.Internal, "gRPC proxy should never reach this stage.")
+}
+
+func (g *grpcProxy) proxyClientToServer(src grpc.ClientStream, dst grpc.ServerStream) chan error {
+	ret := make(chan error, 1)
+	go func() {
+		f := &emptypb.Empty{}
```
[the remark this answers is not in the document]
gmichelo: No, the content goes into the `Empty.unknownFields`.

internal/cli/cmd/cluster/grpcproxy.go
```diff
@@ -0,0 +1,199 @@
...
+		first := true
+		for {
```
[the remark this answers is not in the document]
gmichelo: I find it a bit more confusing that the current state TBH.

n-g (approved):

internal/cli/cmd/cluster/buildproxy.go
```diff
@@ -181,8 +182,9 @@ func (bp *buildProxy) Cleanup() error {
 }
 
 func (bp *buildProxy) Serve(ctx context.Context) error {
-	if err := serveProxy(ctx, bp.listener, func(ctx context.Context) (net.Conn, error) {
-		return bp.instance.NewConn(ctx)
+	sink := tasks.SinkFrom(ctx)
```
hugosantos, line 185: Can we put this behind a flag so we can test the new implementation in staging for a bit?

internal/cli/cmd/cluster/grpcproxy.go
```diff
@@ -0,0 +1,208 @@
...
+				}
+			}
+
+			if err := dst.SendMsg(f); err != nil {
+				ret <- err
+				break
+			}
+		}
+	}()
+	return ret
```
hugosantos, line 187: There's a chance there's a path added above where we'll forget to write `err`; my advice would be to decouple the writing to the channel, from the error handling. Something like:

```
ret := make(chan error, 1)
go func() {
  if err := doProxy(src, dst); err != nil {
    ret <- err
  }
}()
return ret
```

And then have `doProxy` be a func that returns errors.

internal/cli/cmd/cluster/grpcproxy.go
```diff
@@ -0,0 +1,208 @@
...
+		}
+	}
+
+	return status.Errorf(codes.Internal, "gRPC proxy should never reach this stage.")
+}
+
+func (g *grpcProxy) proxyClientToServer(src grpc.ClientStream, dst grpc.ServerStream) chan error {
+	ret := make(chan error, 1)
+	go func() {
+		// Not really empty message. All fields are unmarshalled in Empty.unknownFields.
```
hugosantos, line 157: Ideally we'd close `ret` when the goroutine returns, to avoid leaking a channel.

```
go func() {
  defer close(ret)
}()
```

internal/cli/cmd/cluster/grpcproxy.go
```diff
@@ -0,0 +1,208 @@
...
+		fmt.Fprintf(console.Debug(ctx), "failed to create client stream: %v\n", err)
+		return status.Errorf(codes.Internal, "failed create client: %v", err)
+	}
+
+	s2cErrChan := g.proxyServerToClient(serverStream, clientStream)
+	c2sErrChan := g.proxyClientToServer(clientStream, serverStream)
+	// Make sure to close both client and server connections
+	for i := 0; i < 2; i++ {
+		select {
+		case s2cErr := <-s2cErrChan:
```
hugosantos, line 125: if the channel is closed, s2cErr will be read as `nil` -- I guess that's ok?

If detecting a closed channel is important, then you can do `case s2cErr, ok := <-s2ErrChan:`

hugosantos (approved):

internal/cli/cmd/cluster/grpcproxy.go
gmichelo: The problem with closed channels is that they don't block the `select`, so I need to set them to `nil`. See new commit.

PR #1210, opened 2023-10-24
ensure we report SQL errors on db.Query

n-g:
Fixes NSL-1985

### DIFF
--- a/universe/db/postgres/rows.go
+++ b/universe/db/postgres/rows.go
@@ -35,13 +35,16 @@ type tracedRows struct {
 }
 
 func (r tracedRows) Close() {
-	r.rows.Close()
+	// Ensure that any error observed while reading rows is reported to a span.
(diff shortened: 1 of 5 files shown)


universe/db/postgres/rows.go
```diff
@@ -0,0 +1,73 @@
...
+	sql  string
+	rows pgx.Rows
+}
+
+func (r tracedRows) Close() {
+	r.rows.Close()
+}
+
+func (r tracedRows) Err() error {
+	return withSpan(r.ctx, r.opts, "db.Query.Err", r.sql, func(context.Context) error {
```
nichtverstehen, line 42: It think it's more natural to put this into `Close()`.

`Err()` should be a pure read operation. `Close()` is the actual event.

nichtverstehen (approved):
See the comment.

PR #1211, opened 2023-10-26
buildx: allow logging by default to rotated files

gmichelo:
Part of: NSL-2113


internal/console/logger.go
```diff
@@ -0,0 +1,41 @@
...
+		Filename:   filePath,
+		MaxSize:    10, // megabytes
+		MaxBackups: 3,
+		MaxAge:     10,   // days
+		Compress:   true, // disabled by default
+	}
+
+	return &rotatedDebugLogger{
+		rotatedFile: rotatedFile,
+		logger:      log.New(rotatedFile, "", log.LstdFlags|log.Lmicroseconds),
```
hugosantos, line 30: This doesn't feel quite right; the interface is a writer, but then we treat it as a log with headers etc. I think it would be best to remove this and just write directly to the file; let the caller prepend headers as needed.

internal/workspace/dirs/dirs.go
```diff
@@ -82,6 +82,15 @@ func Subdir(rel string) (string, error) {
 	return filepath.Join(cacheDir, rel), nil
 }
 
+func Log(rel string) (string, error) {
```
hugosantos, line 85: s/Log/Logs/

n-g (approved):

PR #1216, opened 2023-12-04
Support specifying additional capabilities for containers

nichtverstehen:


hugosantos (approved):
Can you emit something to console.Debug that calls that this downgrade is happening?

Also please bump the internal version as we’ll need to depend on it in internal.

nichtverstehen:
Changed this PR significantly. Added capabilities field to `security:`. PTAL

hugosantos (approved):

PR #1226, opened 2024-04-12
providers: add vault secrets provider for TLS certificate

edganiukov:
Fixes NSL-3119


internal/providers/vault/certificate.go
```diff
@@ -0,0 +1,133 @@
...
+			ca := cfg.GetCa()
+			if ca == nil {
+				return nil, fnerrors.BadInputError("invalid vault certificate configuration: missing CA configuration")
+			}
+
+			if err := p.Login(ctx, ca, VaultJwtAudience); err != nil {
+				return nil, err
+			}
+
+			commonName := fmt.Sprintf("%s.%s", strings.ReplaceAll(srvRef.RelPath, "/", "-"), cfg.BaseDomain)
```
n-g, line 43: what about multi-repo?
Rel path is not enough as a unique identifier.

internal/providers/vault/certificate.go
```diff
@@ -0,0 +1,133 @@
...
+		}
+	})
+
+	return rErr
+}
+
+func (p *provider) IssueCertificate(ctx context.Context, ca *vault.CertificateAuthority, cn string) ([]byte, error) {
+	return tasks.Return(ctx, tasks.Action("vault.issue-certificate").Arg("issuer", ca.GetIssuer()).Arg("common-name", cn),
+		func(ctx context.Context) ([]byte, error) {
+			pkiMount, role, ok := strings.Cut(ca.GetIssuer(), "/")
```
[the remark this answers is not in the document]
attilaolah: That could be because strings.Cut is somewhat recent (go1.18). I like Cut better since it always returns three values, with SplitN it is easy to forget about the edge case (CA contains no "/") and index into the result which will panic.

internal/providers/vault/certificate.go
```diff
@@ -0,0 +1,133 @@
...
+const (
+	VaultJwtAudience    = "vault.namespace.systems"
+	vaultRequestTimeout = 10 * time.Second
+)
+
+func Register() {
+	p := &provider{}
+	combined.RegisterSecretsProvider(
+		func(ctx context.Context, srvRef *secrets.ServerRef, cfg *vault.Certificate) ([]byte, error) {
+			ca := cfg.GetCa()
```
attilaolah, line 34: The config is proto3 so you can just use cfg.Ca directly, here and below.

internal/providers/vault/certificate.go
```diff
@@ -0,0 +1,133 @@
...
+
+				p.vaultClient.SetNamespace(caCfg.VaultNamespace)
+
+				idTokenResp, err := fnapi.IssueIdToken(ctx, audience, 1)
+				if err != nil {
+					return err
+				}
+
+				loginResp, err := p.vaultClient.Auth.JwtLogin(ctx, schema.JwtLoginRequest{Jwt: idTokenResp.IdToken},
+					vaultclient.WithMountPath(caCfg.GetAuthMethod()),
```
attilaolah, line 82: Why not just caCfg.AuthMethod?

n-g: We usually prefer the proto accessor methods due to their nil safety.

internal/providers/vault/certificate.go
edganiukov: tbh, there is no strong requirement of having unique server name here, but rather unique CN per region, which ensured with combination of server name + `base_domain`.
Ideally I want to to read server name (`name` in `server` spec) using server ref, but not sure if possible.

internal/providers/vault/certificate.go
edganiukov: same as before - we prefer to use methods to avoid unnecessary panics.

internal/providers/vault/certificate.go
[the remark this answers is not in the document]
edganiukov: +1 to @attilaolah , using `strings.Cut` is safer.

internal/providers/vault/certificate.go
```diff
@@ -0,0 +1,133 @@
...
+	vaultClient *vaultclient.Client
+	once        sync.Once
+}
+
+func (p *provider) Login(ctx context.Context, caCfg *vault.CertificateAuthority, audience string) error {
+	var rErr error
+	p.once.Do(func() {
+		if err := tasks.Return0(ctx, tasks.Action("vault.login").Arg("namespace", caCfg.VaultNamespace).Arg("address", caCfg.VaultAddr),
+			func(ctx context.Context) error {
+				var err error
```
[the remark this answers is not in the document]
edganiukov: changed

internal/providers/vault/certificate.go
```diff
@@ -0,0 +1,133 @@
...
+}
+
+type provider struct {
+	vaultClient *vaultclient.Client
+	once        sync.Once
+}
+
+func (p *provider) Login(ctx context.Context, caCfg *vault.CertificateAuthority, audience string) error {
+	var rErr error
+	p.once.Do(func() {
```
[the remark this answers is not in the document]
edganiukov: Hm, good point. Changed.

universe/vault/types.proto
```diff
@@ -0,0 +1,22 @@
...
+option go_package = "namespacelabs.dev/foundation/universe/vault";
+
+message Certificate {
+    string               base_domain = 1;
+    string               region      = 2;
+    CertificateAuthority ca          = 3;
+}
+
+message CertificateAuthority {
+    string vault_addr      = 1;
```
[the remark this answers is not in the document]
edganiukov: No strong opinion, but prefer not have have lots of nested objects.
What would be a benefit of having `vault` submessage.

framework/secrets/combined/combined.go
```diff
@@ -77,14 +83,14 @@ func NewCombinedSecrets(env cfg.Context) (secrets.SecretsSource, error) {
 	return &combinedSecrets{
 		bindings: bindings,
 		local:    local,
-		loaded:   map[string][]byte{},
-		loading:  map[string]*loadingSecret{},
+		loaded:   map[secretIdentifier][]byte{},
```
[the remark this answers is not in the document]
n-g: Yes, it is about performance. 

Next PR is fine, but FWIW I think the change is not too complicated.
We'd need to define a provider interface and update `RegisterSecretsProvider` to accept that interface.
We can skip `secretIdentifier` then.

internal/providers/vault/certificate.go
```diff
@@ -0,0 +1,148 @@
...
+type provider struct {
+	mtx          sync.Mutex
+	vaultClients map[vaultAuth]*vaultclient.Client
+}
+
+func (p *provider) Login(ctx context.Context, caCfg *vault.CertificateAuthority, audience string) (*vaultclient.Client, error) {
+	p.mtx.Lock()
+	defer p.mtx.Unlock()
+
+	client, ok := p.vaultClients[vaultAuth{caCfg.GetVaultAddr(), caCfg.GetAuthMethod()}]
```
n-g, line 72: Don't we need the namespace to be part of the key?

universe/vault/types.proto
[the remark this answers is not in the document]
n-g: With fresh mind, let's skip the sub message, since `vault` is already in the package.

Let's spell out `address` instead of `addr` for consistency with other protos.

internal/providers/vault/certificate.go
```diff
@@ -0,0 +1,148 @@
...
+	)
+}
+
+type vaultAuth struct {
+	addr       string
+	authMethod string
+}
+
+type tlsKeyPair struct {
+	PrivateKey  string   `json:"private_key"`
```
attilaolah, line 58: One minor annoyance with these values is that it isn't immediately obvious whether the data is in PEM or DER format. I happen to remember that Vault returns them in PEM format, but there's nothing here that hints to that. It'd be nice if these fields would be more explicit, e.g. private_key_pem instead of private_key, or at least that should be documented somehow, maybe in a comment above the field.

internal/providers/vault/certificate.go
```diff
@@ -0,0 +1,148 @@
...
+		},
+	)
+}
+
+type vaultAuth struct {
+	addr       string
+	authMethod string
+}
+
+type tlsKeyPair struct {
```
attilaolah, line 57: Any consumer would need to unmarshal this struct, would it make sense to make it public so I wouldn't need to write my own struct for unmarshaling?

internal/providers/vault/certificate.go
edganiukov: Added

universe/vault/types.proto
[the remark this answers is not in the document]
edganiukov: changed

internal/providers/vault/certificate.go
edganiukov: Changed, now each field has suffix `_pem`

internal/providers/vault/certificate.go
edganiukov: Done. Renamed to `TlsBundle` also

framework/secrets/combined/combined.go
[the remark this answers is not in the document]
edganiukov: The problem is that `Load` function is generic and it adds a headache to use interface with generics.
I will leave as it is in this PR.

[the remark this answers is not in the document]
edganiukov: The problem is that Load function is generic and it adds a headache to use interface with generics.
I will leave as it is in this PR.

n-g (approved):

PR #1227, opened 2024-04-18
vault: Expose TlsBundle type with a few convenience methods

attilaolah:
- Move the type out of the `internal` sub-package so others can depend on it.
- Make JSON encoding/decoding happen in the same package, so changing the format could be done easily.
- Add methods for parsing the TLS certificate and the CA pool from the config.
- Add methods for crafting minimal functional mTLS configs for clients & servers.


universe/vault/bundle.go
```diff
@@ -0,0 +1,70 @@
...
+	CertificatePem string   `json:"certificate_pem"`
+	CaChainPem     []string `json:"ca_chain_pem"`
+}
+
+func Parse(data []byte) (*TlsBundle, error) {
+	tb := TlsBundle{}
+	return &tb, json.Unmarshal(data, &tb)
+}
+
+func ParseFromEnv() (*TlsBundle, error) {
```
edganiukov, line 27: I would let user to provide an env key here.

universe/vault/bundle_test.go
```diff
@@ -0,0 +1,106 @@
...
+func testBundle(t *testing.T) *vault.TlsBundle {
+	tb, err := vault.Parse(testData(t))
+	if err != nil {
+		t.Fatalf("could not parse bundle: %v", err)
+	}
+	return tb
+}
+
+func testData(t *testing.T) []byte {
+	const path = "../../internal/testdata/vault/tls_bundle.json"
```
edganiukov, line 100: I think you can use `const path = "tls_bundle.json"` and it should work

attilaolah: It didn't work for me.

```
    bundle_test.go:103: could not parse "tls_bundle.json": open tls_bundle.json: no such file or directory
```

edganiukov: my bad, missed that it is another directory

attilaolah: Added to `embed.FS` now and moved the `testdata` dir closer, so it looks sane now.

n-g (approved):

PR #1229, opened 2024-04-22
vault/certificate: add an option to provide cn and sans

edganiukov:

PR #1230, opened 2024-04-24
vault: create approle secret provider

attilaolah:


attilaolah:
Hold on, this is missing the mount path, which needs to be different for the app role (compared to the one used locally).

universe/vault/types.proto
```diff
@@ -8,14 +8,22 @@ package foundation.universe.vault;
...
 
-message CertificateAuthority {
-    string vault_address   = 1;
-    string vault_namespace = 2;
-    string issuer          = 3;
-    string auth_method     = 4;
+message VaultProvider {
+    string address = 1;
+    string namespace = 2;
+    string auth_method = 4;
```
edganiukov, line 28:
```suggestion
    string auth_method = 3;
```
As this is a new message.

internal/providers/vault/issue_257.go
```diff
@@ -0,0 +1,54 @@
...
+	return vault.WithHTTPClient(c)
+}
+
+type fix257 struct {
+	rt http.RoundTripper
+}
+
+func (w fix257) RoundTrip(req *http.Request) (*http.Response, error) {
+	res, err := w.rt.RoundTrip(req)
+	if !strings.HasSuffix(req.URL.Path, "/secret-id") ||
```
edganiukov, line 32: This if looks a bit complex (I would split it to multiple if), but as it is a temporary workaround might be OK.

internal/providers/vault/approle.go
```diff
@@ -0,0 +1,75 @@
...
+func (p *provider) CreateSecretId(ctx context.Context, vaultClient *vaultclient.Client, cfg *vault.AppRole) ([]byte, error) {
+	return tasks.Return(ctx, tasks.Action("vault.create-secret-id").Arg("name", cfg.GetName()),
+		func(ctx context.Context) ([]byte, error) {
+			creds := vault.Credentials{
+				VaultAddress:   cfg.Provider.GetAddress(),
+				VaultNamespace: cfg.Provider.GetNamespace(),
+			}
+			wmp := vaultclient.WithMountPath(cfg.GetAuthMethod())
+
+			g := errgroup.Group{}
```
hugosantos, line 42: Please use errgroup.WithContext to respect cancelation.

Or better yet, use foundation's executor.New()

internal/providers/vault/certificate.go
```diff
@@ -26,30 +23,23 @@ const (
...
-			}
+	if secretId.ServerRef == nil {
+		return nil, fnerrors.BadDataError("required server reference is not set")
+	}
 
-			commonName := fmt.Sprintf("%s.%s", strings.ReplaceAll(secretId.ServerRef.RelPath, "/", "-"), cfg.GetBaseDomain())
-			return p.IssueCertificate(ctx, vaultClient, ca.GetIssuer(), commonName)
-		},
-	)
+	commonName := fmt.Sprintf("%s.%s", strings.ReplaceAll(secretId.ServerRef.RelPath, "/", "-"), cfg.GetBaseDomain())
```
hugosantos, line 41: I know you're just refactoring but the cname generation here needs revisiting; we don't use server references as stable identifiers as they depend on the location in the codebase.

We'd use server ids for that.

But those hard to digest, so I'd be inclined to receive the cname explicitly as an argument, instead of generating it.

internal/providers/vault/issue_257.go
hugosantos: It also has a bug as it doesn't always close res.Body. :-)

internal/providers/vault/issue_257.go
```diff
@@ -0,0 +1,54 @@
...
+		err != nil ||
+		res == nil ||
+		res.StatusCode != http.StatusOK ||
+		strings.Split(res.Header.Get("content-type"), ";")[0] != "application/json" {
+		return res, err
+	}
+	defer res.Body.Close()
+
+	body := map[string]any{}
+	json.NewDecoder(res.Body).Decode(&body)
```
hugosantos, line 42: error checking?

If you want to drop errors, please do so explicitly, e.g.:

```
// Swallow the decoding error as we interpret it as a missing data field.
_ = json.NewDecoder(...).Decode(...)
```

internal/providers/vault/issue_257.go
```diff
@@ -0,0 +1,54 @@
...
+
+	if data, ok := body["data"].(map[string]any); ok {
+		if ttl, ok := data["secret_id_ttl"].(float64); ok {
+			data["secret_id_ttl"] = strconv.Itoa(int(ttl))
+		}
+	}
+
+	raw, _ := json.Marshal(body)
```
hugosantos, line 50: Sames.

internal/providers/vault/issue_257.go
```diff
@@ -0,0 +1,54 @@
...
+		return res, err
+	}
+	defer res.Body.Close()
+
+	body := map[string]any{}
+	json.NewDecoder(res.Body).Decode(&body)
+
+	if data, ok := body["data"].(map[string]any); ok {
+		if ttl, ok := data["secret_id_ttl"].(float64); ok {
+			data["secret_id_ttl"] = strconv.Itoa(int(ttl))
```
hugosantos, line 46: This monkey-patching of the response is odd; is this just so you use a type you already have? Why not adjust the type instead to accept a float64 instead?

(If this is indeed needed, would be good to leave why as a comment).

internal/providers/vault/provider.go
```diff
@@ -0,0 +1,87 @@
...
+	combined.RegisterSecretsProvider(p.certificateProvider)
+}
+
+type provider struct {
+	mtx          sync.Mutex
+	vaultClients map[string]*vaultclient.Client
+}
+
+func (p *provider) Login(ctx context.Context, caCfg *vault.VaultProvider, audience string) (*vaultclient.Client, error) {
+	vKey := vaultIdentifier{
```
hugosantos, line 35: Nit: `key` would have been unambiguous enough. :-P

universe/vault/types.proto
attilaolah: Ah right. I renamed an old message, but updated the field number in the parent, so it is indeed a new message now.

internal/providers/vault/provider.go
```diff
@@ -0,0 +1,87 @@
...
+		address:    caCfg.GetAddress(),
+		namespace:  caCfg.GetNamespace(),
+		authMethod: caCfg.GetAuthMethod(),
+	}
+	p.mtx.Lock()
```
hugosantos, line 40: a Mutex is not cancelation aware. Which is why we rarely use mutexes (or we don't do expensive work under a lock).

We have a facility for this already, it's a `tcache`, you can use that instead.

internal/providers/vault/approle.go
```diff
@@ -0,0 +1,75 @@
...
+	if vp == nil {
+		return nil, fnerrors.BadInputError("invalid vault app role configuration: missing provider configuration")
+	}
+
+	vaultClient, err := p.Login(ctx, vp, vaultJwtAudience)
+	if err != nil {
+		return nil, err
+	}
+
+	if secretId.ServerRef == nil {
```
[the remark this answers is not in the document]
attilaolah: Nah, this is leftover copy/pasta.

internal/providers/vault/certificate.go
edganiukov: I actually added this [here](https://github.com/namespacelabs/foundation/pull/1229)
Initially I was thinking to have less secrets defined in `fn-workspace.cue`, but now I also on the side to provide CN explicitly.
I will fix this.

internal/providers/vault/issue_257.go
attilaolah: RoundTrip should not error based on the content… But yes, we should swallow the error, but this isn't good as-is, because in this case we would consume the content.

Let me change this.

internal/providers/vault/certificate.go
attilaolah: OK I'll leave it as-is for now, but then it would be cool if we maybe _required_ the user to specify the cname (or… default tot he server ID unless one is specified?)

internal/providers/vault/provider.go
attilaolah: Indeed! This file is copy/paste though, but Git doesn't seem to have recognised that.

internal/providers/vault/certificate.go
edganiukov: server ID could be an option (e.g. the one defined in `server.cue`), but not sure if it easy to get it at this phase (but didn't spend much time trying to make it work).

internal/providers/vault/approle.go
attilaolah: image

Much better.

internal/providers/vault/provider.go
attilaolah: This too is copy/pasta, but let me learn about tcache.

attilaolah: PTAL, there is some tcache/task nesting that I'm not sure of, but it seems to work (I see the task message showing up during login… it seems to work.)

internal/providers/vault/certificate.go
attilaolah: My suggestion would be: keep it as you added already, since it works, and then switch to using server-id in a follow-up change, and then merge your PR that adds custom CN.

internal/providers/vault/issue_257.go
```diff
@@ -0,0 +1,67 @@
...
+
+func (w fix257) RoundTrip(req *http.Request) (*http.Response, error) {
+	if !strings.HasSuffix(req.URL.Path, "/secret-id") {
+		return w.rt.RoundTrip(req)
+	}
+	res, err := w.rt.RoundTrip(req)
+	if err != nil {
+		return res, err
+	}
+	defer res.Body.Close()
```
attilaolah, line 38: This is actually not correct: if we don't consume and replace the body, we shouldn't close it. It was in fact correct earlier when it wasn't closed after that long if statement :P

internal/providers/vault/provider.go
```diff
@@ -0,0 +1,65 @@
...
+var clients = tcache.NewCache[*vaultclient.Client]()
+
+func Register() {
+	combined.RegisterSecretsProvider(appRoleProvider)
+	combined.RegisterSecretsProvider(certificateProvider)
+}
+
+func login(ctx context.Context, caCfg *vault.VaultProvider, audience string) (*vaultclient.Client, error) {
+	key := caCfg.GetAddress() + caCfg.GetNamespace() + caCfg.GetAuthMount()
+	return clients.Compute(key, func() (*vaultclient.Client, error) {
```
edganiukov, line 30: The usage of mutex would guarantee that the very first login would block all other for the same vault.
Would this be a case here?

edganiukov:
btw, please bum the API version.

internal/providers/vault/provider.go
attilaolah: Yes, apparently the first thing that Compute does is [lock its own mutex](https://github.com/namespacelabs/foundation/blob/ba4497c865c56bc1c98bfb9768127cfed670b032/internal/tcache/tcache.go#L38-L50).

internal/providers/vault/issue_257.go
attilaolah: Added a comment to make the intention clear here.

internal/providers/vault/issue_257.go
attilaolah: Coming back to this, it was in fact intentional that the body isn't closed, since we don't consume the bady if we return early (the caller will consume & close it). But it was also confusing… So I added a comment now and moved the close right next to where we consume the body, to make the intention obvious here.

edganiukov (approved):

PR #1232, opened 2024-04-25
vault: renew tokens

attilaolah:
The Vault client currently does not support automatic token renewal. This is a simple approach that keeps renewing the token until the original context is cancelled.

Some things to consider:

- Add retry. With backoff.
- Tokens have a max-TTL, give up once we reach it.
- The renew request might return a different auth context, with a different lease duration. This could happen if e.g. the max-TTL is 1.5h, and the initial token was for 1h.
- Have some way to tell the consumer that the client token has expired, and there's no way to renew it. In this case, the client should attempt to reconnect.
  - The Secret ID may or may not allow issuing more than one token.
  - If it allows, reconnect with the same secret ID. If not, then reconnecting will fail immediately.

### DIFF
--- a/universe/vault/credentials.go
+++ b/universe/vault/credentials.go
@@ -17,7 +17,7 @@ import (
 )
 
 // Force re-authentication if the token expires in less than this much time.
-const ttlBuffer = time.Second * 8
+const ttlBuffer = time.Minute
 
 type Credentials struct {
 	RoleId   string `json:"role_id"`
@@ -73,9 +73,12 @@ func (h *ClientHandle) Get(ctx context.Context) (*vault.Client, error) {
 	h.m.Lock()
 	defer h.m.Unlock()
 
-	// Re-authenticate if less than six seconds left.
-	if time.Now().Add(ttlBuffer).After(h.expires()) {
-		return h.client, h.renew(ctx)
+	if time.Until(h.expires()) > ttlBuffer {
+		return h.client, nil
+	}
+
+	if err := h.renew(ctx); err != nil {
+		return nil, err
 	}
 
 	return h.client, nil
@@ -92,6 +95,7 @@ func (h *ClientHandle) authenticate(ctx context.Context) error {
 	if err != nil {
 		return err
 	}
+
 	h.auth = resp.Auth
 	h.leased = time.Now()
 	zerolog.Ctx(ctx).Debug().Dur("lease_duration", h.ttl()).Msg("vault: authenticated")
@@ -111,6 +115,7 @@ func (h *ClientHandle) renew(ctx context.Context) error {
 		zerolog.Ctx(ctx).Warn().Msg("vault: token renewal failed, forcing re-auth")
 		return h.authenticate(ctx)
 	}
+
 	h.auth = res.Auth
 	zerolog.Ctx(ctx).Debug().Dur("lease_duration", h.ttl()).Msg("vault: token renewed")
 	h.leased = time.Now()
@@ -121,6 +126,7 @@ func (h *ClientHandle) ttl() time.Duration {
 	if h.auth == nil {
 		return 0
 	}
+
 	return time.Duration(h.auth.LeaseDuration) * time.Second
 }
 
@@ -128,5 +134,6 @@ func (h *ClientHandle) expires() time.Time {
 	if h.auth == nil {
 		return time.Time{}
 	}
+
 	return h.leased.Add(h.ttl())
 }
--- a/universe/vault/credentials.go
+++ b/universe/vault/credentials.go
@@ -99,14 +99,13 @@ func (h *ClientHandle) renew(ctx context.Context) error {
 		return h.authenticate(ctx)
 	}
 
-	// The Vault client library already handles retries, so if renewing the
-	// token fails, we assume it can no longer be renewed. This can happen if
-	// the token was revoked, or if it reached its maximum TTL.
-	h.auth = nil // force re-auth on next try
-
 	res, err := h.client.Auth.TokenRenewSelf(ctx, schema.TokenRenewSelfRequest{})
 	if err != nil {
-		return err
+		// The Vault client library already handles retries, so if renewing the
+		// token fails, we assume it can no longer be renewed. This can happen if
+		// the token was revoked, or if it reached its maximum TTL.
+		zerolog.Ctx(ctx).Warn().Msg("vault: token renewal failed, forcing re-auth")
+		return h.authenticate(ctx)
 	}
 	h.auth = res.Auth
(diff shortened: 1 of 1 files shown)


universe/vault/credentials.go
```diff
@@ -62,5 +64,39 @@ func (c *Credentials) Login(ctx context.Context, options ...vault.ClientOption)
...
+		zerolog.Ctx(ctx).Warn().Msgf("vault: not renewing token, lease too short: %s", lease)
+		return
+	}
+
+	ticker := time.NewTicker(interval)
+	defer ticker.Stop()
+	for {
+		select {
+		case <-ticker.C:
+			if _, err := client.Auth.TokenRenewSelf(ctx, schema.TokenRenewSelfRequest{}); err != nil {
```
n-g, line 93: Ideally, we'd check for the new lease duration. And if it is too short, re-auth.

universe/vault/credentials.go
```diff
@@ -62,5 +64,39 @@ func (c *Credentials) Login(ctx context.Context, options ...vault.ClientOption)
...
+				// TODO: Let the consumer know, so it could try and reconnect if needed?
+				zerolog.Ctx(ctx).Error().Msgf("vault: failed to renew token: %v", err)
+				return // TODO: retry, with backoff
```
n-g, line 96: constant backoff should be fine here https://github.com/namespacelabs/foundation/blob/7223a6e741b1e532bcb4c44eb36edde7b732e9ae/internal/providers/nscloud/ctl/wait.go#L107-L108

universe/vault/credentials.go
```diff
@@ -62,5 +64,39 @@ func (c *Credentials) Login(ctx context.Context, options ...vault.ClientOption)
...
+		return // token does not expire
+	}
+	if !auth.Renewable {
+		zerolog.Ctx(ctx).Warn().Msgf("vault: non-renewable token expires in %s", lease)
+		return
+	}
+
+	interval := lease - time.Minute*2
+	if interval < 0 {
+		zerolog.Ctx(ctx).Warn().Msgf("vault: not renewing token, lease too short: %s", lease)
```
n-g, line 84: Reauth instead?

universe/vault/credentials.go
```diff
@@ -62,5 +64,39 @@ func (c *Credentials) Login(ctx context.Context, options ...vault.ClientOption)
...
 	return client, nil
 }
+
+func renew(ctx context.Context, client *vault.Client, auth *vault.ResponseAuth) {
+	lease := time.Duration(auth.LeaseDuration) * time.Second
+	if lease <= 0 {
+		return // token does not expire
+	}
+	if !auth.Renewable {
+		zerolog.Ctx(ctx).Warn().Msgf("vault: non-renewable token expires in %s", lease)
```
n-g, line 78: schedule a reauth after expiration?

universe/vault/credentials.go
attilaolah: Turns out the Vault client library [already handles retries](https://github.com/hashicorp/vault-client-go/blob/cfdcbeda2e34770a045cc9f71ad25db195e97181/client_option.go#L65) so let's not retry here.

attilaolah:
PTAL. Now there is a client handle that sadly has to contain a mutex to avoid racing reads vs. writes. But the upside is there is no busywork.

If a user still wants to keep the token alive, and thus avoid the latency penalty that comes with having to renew an expired token, they could simply spawn a goroutine that periodically calls `.Get()`, e.g. once a minute.

We could also add a function to the handle to keep it alive in a more efficient way, e.g. by calling `go handle.Keepalive(ctx)` which would block on a ticker + `ctx.Done()`, but I don't plan on using that, so whoever needs it can still add it.

universe/vault/credentials.go
```diff
@@ -48,19 +63,70 @@ func (c *Credentials) Login(ctx context.Context, options ...vault.ClientOption)
...
+}
+
+func (h *ClientHandle) ttl() time.Duration {
+	if h.auth == nil {
+		return 0
+	}
+	return time.Duration(h.auth.LeaseDuration) * time.Second
+}
+
+func (h *ClientHandle) expires() time.Time {
```
n-g, line 127: Prefer

```
	if h.auth == nil {
		return time.Time{}
	}

	return h.leased.Add(h.ttl())
```
(new line after block)

Here, and in other places :)
https://www.notion.so/namespacelabs/Style-Guide-331d14027ec64eb0a4bb4a0a3aa4fd07?pvs=4#3355592134f14a9697eedaf1bc35f599

universe/vault/credentials.go
```diff
@@ -8,11 +8,17 @@ import (
...
+	"sync"
+	"time"
 
 	"github.com/hashicorp/vault-client-go"
 	"github.com/hashicorp/vault-client-go/schema"
+	"github.com/rs/zerolog"
 )
 
+// Force re-authentication if the token expires in less than this much time.
+const ttlBuffer = time.Second * 8
```
n-g, line 20: sounds very tight.
What are typical TTLs?
How about `time.Minute` as a buffer?

universe/vault/credentials.go
```diff
@@ -48,19 +63,70 @@ func (c *Credentials) Login(ctx context.Context, options ...vault.ClientOption)
...
+		creds:  c,
+		client: client,
+	}, nil
+}
+
+func (h *ClientHandle) Get(ctx context.Context) (*vault.Client, error) {
+	h.m.Lock()
+	defer h.m.Unlock()
+
+	// Re-authenticate if less than six seconds left.
```
n-g, line 76: Works, but a bit harder to parse.
We usually prefer simple return guards, e.g. 

```
	if time.Until(h.expires()) > ttlBuffer {
		return h.client, nil
	}

	if err := h.renew(ctx); err != nil {
		return nil, err
	}

	return h.client, nil
```

Also, I'd skip the comment (it is already outdated).

n-g (approved):

universe/vault/credentials.go
attilaolah: The default is an hour, so a minute should be fine as well. I don't expect time skew of anything over a second, but it doesn't hurt much to be a bit more conservative.

universe/vault/credentials.go
attilaolah: Oh we have a style guide :slightly_smiling_face: It seems like a `\n\s*}\n\s*[^}\t]` would match pretty much all bad cases, so maybe we could add a presubmit then?

universe/vault/credentials.go
attilaolah: (I thought I removed that comment already.)

Fair point about early-return, plus it is nice to always easier to follow if one of (ptr, err) is a literal nil.

PR #1233, opened 2024-04-26
nix: add clang-format wrapper to flake

attilaolah:


edganiukov (approved):
works for me

attilaolah:
OK, since it doesn't remove any binary, this seems safe. If someone directly exec's into `clang-format` or does something weird, that's on them I guess.

PR #1235, opened 2024-04-29
vault: add fn extension

attilaolah:


universe/vault/wire.go
```diff
@@ -0,0 +1,19 @@
...
+
+package vault
+
+import (
+	"context"
+	"fmt"
+)
+
+func ProvideClientHandle(ctx context.Context, _ *VaultClientArgs) (*ClientHandle, error) {
+	creds, err := ParseCredentialsFromEnv("VAULT_APP_ROLE")
```
n-g, line 13: Let's wire a CUE secret for this which we can then bind in other workspaces.

Eg. https://github.com/namespacelabs/foundation/blob/8487f73581bf786372b9860707e1e1e9007b4748/universe/monitoring/honeycomb/extension.cue#L21-L22

n-g (approved):

universe/vault/wire.go
attilaolah: But it is unclear how one would pass in the app role. If there are multiple servers using this extension, they would likely use different app roles, but if we move the secret here in the extension, then one can provide only a single source for the secret in the `fn-workspace.cue` file.

n-g: The point of extensions is encapsulation (bundling code with configuration).
As you say, the value of the app role depends on the caller, so extensions are not the right tool here.

An alternative to extensions are resources + providers.
These do allow variable inputs, but the purpose of them is to model an object with a lifecycle (e.g. a DB that requires a schema to be applied on init).

In your case, the best solution is to keep it simple IMO and just offer a plain go package.

attilaolah: Dropping this as discussed. ENV var will be bound at the service level, where it makes more sense.

PR #1237, opened 2024-04-30
providers/vault: add provider struct

attilaolah:


universe/vault/credentials.go
```diff
@@ -19,6 +20,13 @@ import (
 // Force re-authentication if the token expires in less than this much time.
 const ttlBuffer = time.Minute
 
+type Provider struct {
```
n-g, line 23: Nit:

I'd order
```
type Provider struct {
	creds *Credentials

	m      sync.Mutex
	conn *ClientHandle
}
```
To suggest what is protected by the mutex.

n-g (approved):

PR #1241, opened 2024-05-02
std/go/grpc: add an option to provide CA cert via ENV var

edganiukov:
Fixes NSL-3256

### DIFF
--- a/std/go/grpc/servercore/mtls.go
+++ b/std/go/grpc/servercore/mtls.go
@@ -34,9 +34,7 @@ func getMtlsConfig() (*tls.Config, error) {
 	}
 
 	pool := x509.NewCertPool()
-	for _, cert := range tb.CaChainPem {
-		pool.AppendCertsFromPEM([]byte(cert))
-	}
+	pool.AppendCertsFromPEM([]byte(os.Getenv("FOUNDATION_GRPCSERVER_CA_CERT")))
 
 	return &tls.Config{
 		Certificates: []tls.Certificate{cert},


std/go/grpc/servercore/mtls.go
```diff
@@ -33,10 +33,9 @@ func getMtlsConfig() (*tls.Config, error) {
 		return nil, err
 	}
 
+	caCertPem := os.Getenv("FOUNDATION_GRPCSERVER_CA_CERT")
 	pool := x509.NewCertPool()
-	for _, cert := range tb.CaChainPem {
-		pool.AppendCertsFromPEM([]byte(cert))
-	}
+	pool.AppendCertsFromPEM([]byte(caCertPem))
```
attilaolah, line 38: I'd just inline it since we're not doing any checks whether the cert is there or not. The env var name makes it clear what we expect to find there.

```go
pool.AppendCertsFromPEM([]byte(os.Getenv("FOUNDATION_GRPCSERVER_CA_CERT")))
```

attilaolah (approved):

std/go/grpc/servercore/mtls.go
edganiukov: done. I do not want to do any checks as its not required to provide a CA cert.

attilaolah (approved):

std/go/grpc/servercore/mtls.go
```diff
@@ -34,9 +34,7 @@ func getMtlsConfig() (*tls.Config, error) {
 	}
 
 	pool := x509.NewCertPool()
-	for _, cert := range tb.CaChainPem {
```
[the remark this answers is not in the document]
edganiukov: I would keep `CaChainPem` just to have the same what Vault returns.
But makes sense to update `universe/vault/bundle.go`. I will do this in a follow-up PR.

PR #1242, opened 2024-05-03
grpc: add an option to provide gRPC dial opts in Loopback func

edganiukov:
So it possible to configure custom credentials, etc.

### DIFF
--- a/std/grpc/wire.go
+++ b/std/grpc/wire.go
@@ -8,6 +8,8 @@ import (
 	"context"
 	"flag"
 	"fmt"
+	"net"
+	"strconv"
 	"strings"
 
 	"google.golang.org/grpc"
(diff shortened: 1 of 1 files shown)


edganiukov:
hm, okay, that actually doesn't work. Looking how to fix.

edganiukov:
For now changing only `Loopback` functions.

std/grpc/wire.go
```diff
@@ -35,9 +35,17 @@ func ProvideConn(ctx context.Context, req *Backend) (*grpc.ClientConn, error) {
 	return client.Dial(ctx, endpoint, grpc.WithTransportCredentials(insecure.NewCredentials())) ///  XXX mTLS etc.
 }
 
-func Loopback(ctx context.Context) (*grpc.ClientConn, error) {
+func Loopback(ctx context.Context, grpcOpts ...grpc.DialOption) (*grpc.ClientConn, error) {
 	endpoint := fmt.Sprintf("127.0.0.1:%d", server.ListenPort())
```
attilaolah, line 39: BTW why don't we use `net.JoinHostPort` for these?

attilaolah (approved):

std/grpc/wire.go
edganiukov: I guess here is not that bad as we explicitly use IPv4 (and  `server.ListenPort()` is an integer and would need to be converted to string to be use in `net.JoinHostPort`).

edganiukov: and from what I heard people do not like `strconv.Itoa()`

edganiukov: changed btw

attilaolah: Yeah it sucks that the port must be a string in `net.JoinHostPort`, I forgot about that.

attilaolah (approved):

PR #1244, opened 2024-05-08
providers/vault: Allow configuring additional sans

attilaolah:


internal/providers/vault/certificate.go
```diff
@@ -54,6 +54,12 @@ func certificateProvider(ctx context.Context, conf cfg.Configuration, secretId s
 	if certConfig, ok := GetCertificateConfig(conf); ok && certConfig.GetBaseDomain() != "" {
```
edganiukov, line 54: I would move 55-56 lines into its own `if`. E.g.:
```
if certConfig, ok := GetCertificateConfig(conf); ok {
    if certConfig.GetBaseDomain() != "" {
	    req.commonName = fmt.Sprintf("%s/%s", certConfig.GetBaseDomain(), req.commonName)
	    req.excludeCnFromSans = true
	}

	for _, domain := range certConfig.GetSansDomains() {
		for _, san := range cfg.GetSans() {
			req.sans = append(req.sans, fmt.Sprintf("%s.%s", san, domain))
		}
	}
}
```

attilaolah: Good point.

edganiukov (approved):

PR #1247, opened 2024-05-16
auth: do an online check for session tokens

attilaolah:
The intention is to check whether the token has been revoked.

### DIFF
--- a/internal/fnapi/tokens.go
+++ b/internal/fnapi/tokens.go
@@ -106,16 +106,18 @@ func ResolveSpec() (string, error) {
 	return "", nil
 }
 
-func VerifySession(ctx context.Context) (bool, error) {
-	req := emptypb.Empty{}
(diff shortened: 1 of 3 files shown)


internal/fnapi/tokens.go
```diff
@@ -103,3 +105,19 @@ func ResolveSpec() (string, error) {
...
+func VerifySession(ctx context.Context, t Token) error {
+	st := ""
+	if t, ok := t.(*auth.Token); ok && t.SessionToken != "" {
+		st = t.SessionToken
+	} else {
+		return errors.New("not a session token")
+	}
+
+	return (Call[*emptypb.Empty]{
+		Method: "nsl.signin.SigninService/VerifySession",
```
gmichelo, line 118: nit: I guess we could have used the Go buf SDK.

gmichelo (approved):

PR #1249, opened 2024-05-17
auth: add hidden check-login --duration command

attilaolah:
Fixes NSL-3354.

### DIFF
--- a/internal/cli/cmd/auth/checklogin.go
+++ b/internal/cli/cmd/auth/checklogin.go
@@ -1,0 +1,53 @@
+// Copyright 2022 Namespace Labs Inc; All rights reserved.
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+
+package auth
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/spf13/cobra"
+	"namespacelabs.dev/foundation/internal/cli/fncobra"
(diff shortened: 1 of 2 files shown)


internal/cli/cmd/auth/checklogin.go
```diff
@@ -0,0 +1,53 @@
...
+	"namespacelabs.dev/foundation/internal/cli/fncobra"
+	"namespacelabs.dev/foundation/internal/fnapi"
+)
+
+func NewCheckLoginCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:    "check-login",
+		Short:  "Return a successful exit code if the caller is still authenticated to Namespace.",
+		Args:   cobra.NoArgs,
+		Hidden: true,
```
gmichelo, line 22: Nit: If customer is going to use this, should we make it public? Perhaps after they confirmed it looks good on their side.

gmichelo (approved):

internal/cli/cmd/auth/checklogin.go
attilaolah: Sounds good to me, although we might as well make it public now, since we'd probably not want to make a backwards-incompatible change even for a hidden command.

But also OK for me to wait until they confirm it works as expected for them, we have many hidden commands so apparently this wouldn't be much of an outlier.

PR #1289, opened 2024-08-15
cmd/gitcheckout: resolve submodule relative URL into absolute one from remote origin

gmichelo:
Part of: NSL-3898

### DIFF
--- a/internal/cli/cmd/cluster/git.go
+++ b/internal/cli/cmd/cluster/git.go
@@ -1,0 +1,73 @@
+package cluster
+
+import (
+	"fmt"
+	"path"
+	"path/filepath"
+	"strings"
+)
+
+type gitScheme string
+
+const (
+	SSH   gitScheme = "ssh"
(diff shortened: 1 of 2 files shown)


internal/cli/cmd/cluster/git.go
```diff
@@ -0,0 +1,73 @@
...
+
+func (g giturl) URL() string {
+	if g.scheme == SSH {
+		return fmt.Sprintf("git@%s:%s/%s.git", g.hostname, g.owner, g.repository)
+	} else {
+		return fmt.Sprintf("https://%s/%s/%s.git", g.hostname, g.owner, g.repository)
+	}
+}
+
+func parseGitUrl(url string) (giturl, error) {
```
pavolmarko, line 30: Probably not relevant in this particular case but I think file:// also exists (which can be implied by being left out) and git://
https://git-scm.com/book/en/v2/Git-on-the-Server-The-Protocols

n-g:
LGTM

PR #1291, opened 2024-08-15
git-checkout: Skip stale .gitmodules entries

pavolmarko:
Only process .gitmodules entries if they point to a "commit" type object.

Do this because regular git submodule commands behave like that.

Bug: NSL-3880


n-g:
LGTM

PR #1310, opened 2024-09-09
integrations/golang: Use full package paths when building.

nichtverstehen:
Carefully tracking the path to the built package within the Go module allows us to refer to binaries in a build plan across modules.

Previously given the following specification in example.com/foo/xxx/binary.cue

    build_plan: [{binary: "example.com/bar/yyy"}]

we would try to build a package with the desired local name (yyy), but within the "calling" module (example.com/foo/yyy).


hugosantos (approved):
If this is backwards compatible LGTM

nichtverstehen:
I've tried build a few binaries in various locations including in workspace submodules and it seems that all works. 🤞

PR #1311, opened 2024-09-09
Unupgrade distribution/v3.

nichtverstehen:
Internal registry is not compatible with the beta version.


n-g:
FOLLOWUP https://github.com/namespacelabs/foundation/pull/1312/files

PR #1313, opened 2024-09-09
Try to satisfy all deps with oras.land/oras-go@v1.2.5.

nichtverstehen:
v1.2.6 requires distribution@beta1, but internal registry is not compatible with it.
v1.2.4 requires older docker@24, but buildkit requires @27.


n-g:
LGTM

PR #1314, opened 2024-09-10
buildkitknob

hugosantos:
- buildkit: add ability to pass a flag to connect to an existing buildkit: --buildkit_addr
- servercore: introduced SharedMtlsGrpcListenerConfiguration that allows creating separate listeners that share the same mtls configuration.


edganiukov (commented):
LGTM

PR #1316, opened 2024-09-10
bazel: Add credential helper

pavolmarko:
Uses Bearer token in header auth.

Bug: NSL-3913


n-g:
FOLLOWUP https://github.com/namespacelabs/foundation/pull/1318/files https://github.com/namespacelabs/foundation/pull/1319/files

PR #1327, opened 2024-09-16
buildkit proxy: only discard and close connections if they're shut down

pavolmarko:
Specifically, don't close when Idle.

Bug: https://linear.app/namespacelabs/issue/NSL-4181/nsc-client-side-proxy-regression


gmichelo (approved):
LGTM

PR #1328, opened 2024-09-17
Ensure that pod observer never acts on stale watch events

n-g:


hugosantos (commented):
LGTM

But I would have lifted pkg/listandwatch from internal.

PR #1332, opened 2024-09-25
gitcheckout: Fix workers exiting when a work item is done

pavolmarko:
Workers should only exit when the full process is done (doneSignal) or
when canceled.
They should not exit when a single work item finished without error.


n-g:
LGTM

PR #1336, opened 2024-10-04
tracing: add http.request.header.referer

tomasz-sodzawiczny:
towards NSL-4308

### DIFF
--- a/std/monitoring/tracing/wire.go
+++ b/std/monitoring/tracing/wire.go
@@ -222,7 +222,17 @@ func Prepare(ctx context.Context, deps ExtensionDeps) error {
 	)
 
 	deps.Middleware.Add(func(h http.Handler) http.Handler {
-		return otelhttp.NewHandler(h, "",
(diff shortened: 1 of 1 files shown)


std/monitoring/tracing/wire.go
```diff
@@ -222,7 +222,17 @@ func Prepare(ctx context.Context, deps ExtensionDeps) error {
 	)
 
 	deps.Middleware.Add(func(h http.Handler) http.Handler {
-		return otelhttp.NewHandler(h, "",
+		return otelhttp.NewHandler(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			// add referer
+			span := t.SpanFromContext(r.Context())
+			if span.IsRecording() {
```
gmichelo, line 228: Nit: `span.IsRecording()` should not be needed. Because if span is not present, then `SetAttributes` is a safe no-op.

gmichelo (approved):
LGTM

PR #1348, opened 2024-11-05
Use session client cert for buildx outside of ns instances

pavolmarko:
If not running on a ns instance and a session token is available, use a session client cert for auth to the buildx proxy.

Also refresh it if it existed and then nsc login was performed.


gmichelo (commented):
LGTM

PR #1352, opened 2024-11-20
replace with new containerd to pull in locking fix and add ability to…

n-g:
… exclude locking on empty layers

Towards NSL-4566


nichtverstehen (commented):
lgtm

PR #1354, opened 2024-11-20
vault: add retries to read-only calls

gmichelo:


internal/providers/vault/secret.go
```diff
@@ -29,34 +30,36 @@ func secretProvider(ctx context.Context, conf cfg.Configuration, secretId secret
...
+				secretPkg, secretKey, found := strings.Cut(secretRef, ":")
+				if !found {
+					return nil, fnerrors.BadInputError("invalid vault secret reference: expects secret refernece in format '<mount>/<path>:<key>'")
+				}
+
+				secretMount, secretPath, found := strings.Cut(secretPkg, "/")
+				if !found {
+					return nil, fnerrors.BadInputError("invalid vault secret package: expects secret package in format '<mount>/<path>'")
+				}
+				vaultClient, err := login(ctx, vaultConfig)
```
edganiukov, line 43: you wrap login call in this function, but not here - https://github.com/namespacelabs/foundation/pull/1354/files#diff-75ea27d66aada54407e9691a93b5e48da8ea56c29a6ae2f852da1bc38102bd9aR27
I am wondering if it makes sense to do retry inside `login` and move it here out of the retry function.

edganiukov (approved):
lgtm

PR #1357, opened 2024-12-06
Set GRPC_ENFORCE_ALPN_ENABLED false for go-grpc servers

pavolmarko:
This is necessary to support some clients with the go-grpc 1.67+ uprev.

See also NSL-4866 and NSL-4857.


n-g (commented):
LGTM

PR #1358, opened 2024-12-11
Add ability to attach secrets with nsc build.

n-g:
Fixes NSL-4899

### DIFF
--- a/internal/cli/cmd/cluster/build.go
+++ b/internal/cli/cmd/cluster/build.go
@@ -14,6 +14,8 @@ import (
 	"strings"
 	"time"
 
+	controllerapi "github.com/docker/buildx/controller/pb"
+	"github.com/docker/buildx/util/buildflags"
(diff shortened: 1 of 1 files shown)


internal/cli/cmd/cluster/build.go
```diff
@@ -380,17 +394,23 @@ func startBuilds(ctx context.Context, contextDir string, fragments []buildFragme
...
 
-	eg.Go(func(_ context.Context) error {
-		<-pw.Done()
-		return pw.Err()
+	eg.Go(func(ctx context.Context) error {
+		select {
+		case <-pw.Done():
+			return pw.Err()
+
+		case <-ctx.Done():
```
gmichelo, line 411: I geuss this is not necessary because:
```
eg := executor.New(ctx, "nsc/build")
```

already is cancelled by `ctx`?

gmichelo (commented):
LGTM

PR #1365, opened 2024-12-30
Don't digest local inputs in buildkit builds by default

pavolmarko:
This is a quick workaround for https://linear.app/namespacelabs/issue/NSL-4847/ns-build-binary-consumes-all-memory


n-g (commented):
LGTM

PR #1370, opened 2025-01-10
Handle db close errors

InverseIntegral:


n-g (approved):
LGTM

Style-wise, I'd prefer

```
defer func() {
			if err := db.Close(); err != nil {
				log.Printf("unable to close database connection: %v", err)
			}
		}()
```

similar below.

PR #1398, opened 2025-02-24
cmd/artifact: Add cache-url command for pull-through caching.

nichtverstehen:
ref NSL-5171

### DIFF
--- a/internal/cli/cmd/cluster/artifact.go
+++ b/internal/cli/cmd/cluster/artifact.go
@@ -6,17 +6,28 @@ package cluster
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"io"
+	"net/http"
 	"os"
+	"time"
 
+	"github.com/cenkalti/backoff"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
 	"namespacelabs.dev/foundation/internal/cli/fncobra"
 	"namespacelabs.dev/foundation/internal/console"
 	"namespacelabs.dev/foundation/internal/fnerrors"
 	"namespacelabs.dev/integrations/api/storage"
 	"namespacelabs.dev/integrations/auth"
+)
+
+const (
+	mainArtifactNamespace = "main"
+	cacheURLRetries       = 3
 )
 
 func NewArtifactCmd() *cobra.Command {
@@ -28,6 +39,7 @@ func NewArtifactCmd() *cobra.Command {
 
 	cmd.AddCommand(newArtifactUploadCmd())
 	cmd.AddCommand(newArtifactDownloadCmd())
+	cmd.AddCommand(newArtifactCacheURLCmd())
 
 	return cmd
 }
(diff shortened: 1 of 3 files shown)


n-g:
Thinking about it more, artifacts is kinda an implementation detail.
Maybe this should be something like `nsc curl <url> -o <file> --cache`

nichtverstehen:
That would suggest that this command fully supports HTTP cache semantics (ETag, Cache-Control, etc). I don't think it's useful to support that, but we shouldn't imply that it's there.

internal/cli/cmd/cluster/artifact.go
```diff
@@ -102,24 +125,160 @@ func newArtifactDownloadCmd() *cobra.Command {
...
+		Use:   "cache-url [target url] [destination filename] { --renew }",
+		Short: "Download an arbitrary URL using a pull-through cache.",
+		Long: `Download the content from an arbitrary URL and cache it for fast access.
+
+If the content is already present in the artifact cache for the given URL it will be used instead.
+
+The content at the URL is assumed to be immutable.`,
+		Args: cobra.ExactArgs(2),
+	}).WithFlags(func(flags *pflag.FlagSet) {
+		flags.BoolVar(&renew, "renew", false, "Force-download from the source and update the cached content.")
```
n-g, line 151: hmm, should we use a staleness limit instead (e.g. --max-age)?

internal/cli/cmd/cluster/artifact.go
```diff
@@ -102,24 +125,160 @@ func newArtifactDownloadCmd() *cobra.Command {
...
+				fmt.Fprintf(console.Stderr(ctx), "Artifact not found in cache; downloading from source...\n")
+				// Fallthrough.
+			}
 		}
 
-		fmt.Fprintf(console.Stdout(ctx), "Downloaded %s (namespace %s) to %s", src, namespace, dest)
+		cachePath := cacheArtifactPath(time.Now(), parsedURL)
+		labels := map[string]string{cacheSourceURLLabel: sourceURL}
+
+		backoff.RetryNotify(func() error {
```
n-g, line 201: I'd say that we are blindly retrying too many errors here.
I would invert the logic and treat all errors as permanent unless it is a known-safe-to-retry error.

internal/cli/cmd/cluster/artifact.go
```diff
@@ -102,24 +125,160 @@ func newArtifactDownloadCmd() *cobra.Command {
...
+	defer w.Close()
+
+	if _, err := io.Copy(w, reader); err != nil {
+		return fnerrors.New("failed to write %q: %w", dest, err)
+	}
+	return nil
+}
+
+func cacheArtifactPath(now time.Time, sourceURL *url.URL) string {
+	safePath := slashRuns.ReplaceAllString(strings.Join(validPath.FindAllString(sourceURL.Path, -1), "-"), "/")
```
n-g, line 275: FindAllString feels wrong here. Wouldn't this match the full path?

n-g (commented):
maybe `nsc cache --remote-url <src> --out <dest>`

internal/cli/cmd/cluster/artifact.go
nichtverstehen: Good idea. Introduced `--max_age` instead of renew.

internal/cli/cmd/cluster/artifact.go
nichtverstehen: Done.

internal/cli/cmd/cluster/artifact.go
```diff
@@ -102,24 +125,160 @@ func newArtifactDownloadCmd() *cobra.Command {
...
+
+	return fncobra.Cmd(&cobra.Command{
+		Use:   "cache-url [target url] [destination filename] { --renew }",
+		Short: "Download an arbitrary URL using a pull-through cache.",
+		Long: `Download the content from an arbitrary URL and cache it for fast access.
+
+If the content is already present in the artifact cache for the given URL it will be used instead.
+
+The content at the URL is assumed to be immutable.`,
+		Args: cobra.ExactArgs(2),
```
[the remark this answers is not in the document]
nichtverstehen: Done.

internal/cli/cmd/cluster/artifact.go
nichtverstehen: It will match all safe parts of the path. Then we'll merge them. In principle we just need to get rid of unsafe characters.

n-g (commented):
LGTM

PR #1408, opened 2025-03-12
Add lock-free ensure that column is nullable

InverseIntegral:


n-g (approved):
LGTM

But use this with care (editing constraints changes the contract, so it may break clients).

PR #1412, opened 2025-03-14
buildx: fix nsc docker buildx {status,cleanup}

pavolmarko:
These now also work with the server-side proxy.

While we're at it, make handling of the state directory a bit less confusing.


gmichelo (commented):
LGTM

PR #1413, opened 2025-03-14
buildkit: use remote build proxy if hinted

pavolmarko:
If the server gives a hint, use the server-side buildkit proxy.


gmichelo (commented):
LGTM

PR #1423, opened 2025-03-31
Allow cleanup of Postgres Helper functions

pavolmarko:


library/oss/postgres/prepare/database/cleanup_functions.sql
```diff
@@ -0,0 +1,7 @@
+-- should clean up all functions that are created by functions.sql
```
n-g, line 1: At least add a comment to functions.sql (to remember updating this).

An alternative would be to define a go struct with turnup or turndown instructions.

n-g (approved):
LGTM

library/oss/postgres/prepare/database/cleanup_functions.sql
pavolmarko: You mean, move this out of a data file and define it in go?

could even be non-ugly with raw string literals

n-g: yep

pavolmarko: done

PR #1427, opened 2025-04-08
buildx: atomic client cred writing

pavolmarko:
Avoids race conditions when multiple builds are started in parallel.

Also avoid writing credentials files at all if instance credentials can be reused.


gmichelo (commented):
LGTM

PR #1442, opened 2025-05-05
nsc/gcp: allow configuring token duration for nsc gcp impersonate

annervisser:
#### nsc/gcp: allow configuring token duration for nsc gcp impersonate

- Add --duration flag to change the namespace token validity duration
- Add --gcp_token_lifetime flag to change configured gcp token duration
- Use existing duration flag for token in aws, ecr, oidc federation cmds


n-g (commented):
LGTM

PR #1486, opened 2025-06-13
build/binary: Stamp built images with build provenance.

nichtverstehen:
Set ns.build-workflow-url and ns.source-commit annotations.

ref NSL-5715


n-g (commented):
LGTM

PR #1487, opened 2025-06-17
feat(vault): add an option to do not prepend base domain to CN

edganiukov:
So that we can have any CNs (mostly to use with external services).


n-g (commented):
LGTM

PR #1698, opened 2026-02-04
Update integrations to enable CreateNewVersionIfExists for artifact uploads

hugosantos:
Updates the integrations dependency to include the change that enables `CreateNewVersionIfExists: true` by default when uploading artifacts via `nsc artifact upload`.

This means artifact uploads will automatically create new versions instead of failing when an artifact already exists.

Also removes the outdated 'single file only' limitation notes from upload/download help text.

Depends on: https://github.com/namespacelabs/integrations/pull/53


InverseIntegral (commented):
LGTM

PR #1785, opened 2026-04-29
servercore: trigger HTTP/2 GOAWAY on lameduck

hugosantos:
## Summary

- `NewHttp2CapableServer` now calls `http2.ConfigureServer` so h2c-hijacked connections register with the `http2.Server`'s internal state and `http.Server.Shutdown` actually fans out GOAWAY frames to them.
- New `LameduckFuncsByName` registry that runs after `MarkShutdownStarted` and before any drain func, used to signal clients we are going away (e.g. via GOAWAY) without waiting for in-flight work.
- `Listen` auto-registers a lameduck for the http server that calls `Shutdown` asynchronously, so SIGTERM immediately initiates the HTTP/2 graceful shutdown handshake while drain funcs continue to hold the process up.

## Validation

- `go build ./...`
- `go vet ./std/go/grpc/... ./std/go/core/...`

### DIFF
--- a/std/go/grpc/drain.go
+++ b/std/go/grpc/drain.go
@@ -11,6 +11,13 @@ import (
 var (
 	DrainFunc        func()
 	DrainFuncsByName = map[string]func(){}
+
+	// LameduckFuncsByName are invoked after MarkShutdownStarted and before any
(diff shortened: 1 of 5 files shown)


InverseIntegral (commented):
LGTM

PR #1790, opened 2026-05-04
feat: nsc instance report for CSVs

nhalford:
This adds support for `nsc instance report` to create CSV output of usage reports. Requires a `--start` timestamp and optional `--end` (defaults to now otherwise).

Currently, filters are not supported, but they will be added as a next step.

### DIFF
--- a/internal/cli/cmd/cluster/generatereport.go
+++ b/internal/cli/cmd/cluster/generatereport.go
@@ -1,0 +1,249 @@
+// Copyright 2022 Namespace Labs Inc; All rights reserved.
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+
+package cluster
+
+import (
+	"context"
+	"encoding/csv"
+	"io"
+	"os"
+	"strconv"
+	"time"
+
+	computev1beta "buf.build/gen/go/namespace/cloud/protocolbuffers/go/proto/namespace/cloud/compute/v1beta"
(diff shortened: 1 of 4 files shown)


internal/cli/cmd/cluster/generatereport.go
```diff
@@ -0,0 +1,138 @@
...
+		defer client.Close()
+
+		req := &computev1beta.GenerateReportRequest {
+			StartTime:	timestamppb.New(startTs),
+			EndTime: 	timestamppb.New(endTs),
+		}
+
+		resp, err := client.Usage.GenerateReport(ctx, req)
+		if err != nil {
+			return err
```
InverseIntegral: Might be nice to have a slightly nicer error here but not super important

internal/cli/cmd/cluster/generatereport.go
```diff
@@ -0,0 +1,138 @@
...
+			}
+		}
+		return nil
+	})
+
+	return cmd
+
+}
+
+func entryToCSVLine(entry *computev1beta.InstanceReportEntry) string {
```
InverseIntegral: You can make this a bit nicer by using a `encoding/csv` `Writer`: https://pkg.go.dev/encoding/csv#Writer

InverseIntegral (commented):
LGTM, just some minor suggestions

InverseIntegral (approved):
LGTM

PR #1800, opened 2026-05-12
Use intstr.Parse for update_strategy max_surge/max_unavailable

hugosantos:
## Summary

- Switches `max_surge` and `max_unavailable` parsing from `intstr.FromString` to `intstr.Parse` in stateless Deployment and DaemonSet rolling update configs.
- Allows integer values (e.g. `"0"`, `"1"`) in addition to percentage strings (e.g. `"25%"`). Previously plain integer strings failed at apply time with: `a valid percent string must be a numeric string followed by an ending '%'`.
- Bumps `api_version` to 160.

## Validation

- `go build ./internal/runtime/kubernetes/...`


annervisser (commented):
LGTM

PR #1855, opened 2026-07-03
base-image upload: add --from-file flag for local tar images

avdv:
Add a --from-file flag to `nsc base-image upload` that loads the source image from a local tar file (docker/podman save format) instead of pulling from a remote registry.

This is useful for images built locally with Podman or Nix that are not pushed to any registry:

    podman save localhost/worker-image:latest -o worker-image.tar
    nsc base-image upload --from-file worker-image.tar worker-image:latest

When --from-file is used, only the target tag argument is required (the source-image argument is omitted). Without --from-file, the original two-argument form is unchanged.

### DIFF
--- a/internal/cli/cmd/cluster/baseimage/upload_base_image.go
+++ b/internal/cli/cmd/cluster/baseimage/upload_base_image.go
@@ -55,6 +55,7 @@ func newUploadBaseImageCmd() *cobra.Command {
 		}
 
 		var img v1.Image
+		var index v1.ImageIndex
 
 		if *fromFile != "" {
 			fmt.Fprintf(console.Info(ctx), "Loading image from file: %s\n", *fromFile)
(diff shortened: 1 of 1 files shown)


internal/cli/cmd/cluster/baseimage/upload_base_image.go
```diff
@@ -80,38 +107,13 @@ func newUploadBaseImageCmd() *cobra.Command {
...
-			annotated := mutate.Annotations(img, annotations).(v1.Image)
-			if h, err = annotated.Digest(); err != nil {
-				return err
-			}
-
-			if err := remote.Write(targetRef, annotated, remoteOpts...); err != nil {
-				return err
-			}
-		default:
-			return fmt.Errorf("unsupported image type: %s", desc.MediaType)
```
avdv, line 114: Note, this switch statement was removed because it only worked for the Image case anyway (since `makeAnnotations(ctx, desc, *annotateWithDigest)` in the original code already called `desc.Image()` unconditionally which I think would return an error for indexes).

avdv: Disregard this, I looked a bit deeper and AFAIU this was working properly before. Adjusted to make it work again.

PR #1858, opened 2026-07-03
cli/cluster: add nsc reservation list/describe/cancel/wait commands

annervisser:
Add `nsc reservation` backed by the public ReservationService, with the list, describe, cancel, and wait subcommands.

- Reservations are served over the Connect/JSON transport (like the private InstanceService), so requests go through `fnapi.Call` with protojson against the regional endpoint.
- `wait` polls `DescribeReservation` until the reservation is fulfilled and then prints the instance id; it supports `--cidfile` and `--wait_timeout`.

The `nsc reservation create` command (and its `nsc instance reserve` alias) is split out into a separate, stacked PR.

Amp-Thread-ID: https://ampcode.com/threads/T-019f1880-ac56-7349-b5df-36566f7fc069

### DIFF
--- a/internal/cli/cmd/cluster/reservation.go
+++ b/internal/cli/cmd/cluster/reservation.go
@@ -1,0 +1,318 @@
+// Copyright 2022 Namespace Labs Inc; All rights reserved.
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+
+package cluster
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io"
+	"os"
+	"strings"
+	"text/tabwriter"
+	"time"
+
+	expcompute "buf.build/gen/go/namespace/cloud/protocolbuffers/go/proto/namespace/experimental/compute"
+	"github.com/spf13/cobra"
+	"google.golang.org/protobuf/encoding/protojson"
+	"google.golang.org/protobuf/proto"
+	"google.golang.org/protobuf/types/known/timestamppb"
+	"namespacelabs.dev/foundation/internal/cli/fncobra"
+	"namespacelabs.dev/foundation/internal/console"
+	"namespacelabs.dev/foundation/internal/fnapi"
+	"namespacelabs.dev/foundation/internal/fnerrors"
+	"namespacelabs.dev/foundation/internal/providers/nscloud/endpoint"
+	"namespacelabs.dev/foundation/std/tasks"
+)
+
+// NewReservationCmd implements `nsc reservation`.
+func NewReservationCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "reservation",
+		Short: "Reservation-related activities.",
+	}
+
+	cmd.AddCommand(NewReservationWaitCmd())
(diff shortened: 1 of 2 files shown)


internal/cli/cmd/cluster/reservation.go
```diff
@@ -0,0 +1,667 @@
...
+	cmd := &cobra.Command{
+		Use:   "list",
+		Short: "Lists your reservations.",
+		Args:  cobra.NoArgs,
+	}
+
+	output := cmd.Flags().StringP("output", "o", "plain", "One of plain or json.")
+	since := fncobra.Duration(cmd.Flags(), "since", 7*24*time.Hour, "Constrain the list to reservations active within this duration. Set to 0 to list all.")
+	maxEntries := cmd.Flags().Int64("max_entries", 100, "Maximum number of reservations to return.")
+	excludeFulfilled := cmd.Flags().Bool("exclude_fulfilled", false, "If true, omit reservations that have already been fulfilled.")
```
nhalford: I suggest we leave this off for now, and we can add the filters properly when we update the API to include string matchers. This flag as is could prove a little unwieldy depending on how we handle that in the cli.

internal/cli/cmd/cluster/reservation.go
```diff
@@ -0,0 +1,667 @@
...
+	}
+
+	switch {
+	case v.InstanceId != "":
+		v.Status = "fulfilled"
+	case deadline != nil && time.Now().After(*deadline):
+		v.Status = "expired"
+	default:
+		v.Status = "pending"
+	}
```
nhalford: The status is already in the API, I think this is unnecessary(?).

internal/cli/cmd/cluster/reservation.go
```diff
@@ -0,0 +1,667 @@
...
+	purpose := cmd.Flags().String("purpose", "Manually reserved from CLI", "What documented purpose to attach to the created instance.")
+	selectors := cmd.Flags().StringSlice("selectors", nil, "Select platform/base image based on specific properties (prop1=value1,prop2=value2).")
+	ingress := cmd.Flags().String("ingress", "", "If set, configures the ingress of this instance. Valid options: wildcard.")
+	userSshKey := cmd.Flags().String("ssh_key", "", "Injects the specified ssh public key in the created instance.")
+	enable := cmd.Flags().StringSlice("enable", nil, "Enable a feature, e.g. --enable=kubernetes:1.33")
+	volumes := cmd.Flags().StringSlice("volume", nil, "Attach a volume to the instance, {cache|persistent}:{tag}:{mountpoint}:{size}")
+	output := cmd.Flags().StringP("output", "o", "plain", "One of plain or json.")
+
+	duration := fncobra.Duration(cmd.Flags(), "duration", 0, "For how long to run the instance once it is created.")
+	reservationDeadline := fncobra.Duration(cmd.Flags(), "reservation_deadline", 15*time.Minute, "For how long the server should keep trying to fulfill the reservation before giving up.")
```
nhalford: Maybe `reservation_duration` is more natural here, since deadline usually refers to an absolute time?

annervisser: reservation_timeout

nhalford (approved):

PR #1866, opened 2026-07-11
nsc: add optional remote execution to bazel setup

hugosantos:
## Summary

- Adds a hidden `nsc bazel setup` command with remote execution enabled by default.
- Supports `--remote=false` to configure remote caching and build events without adding remote execution flags to the generated bazelrc.

## Validation

- `go test ./internal/cli/cmd/cluster -count=1`
- `go build -o /tmp/nsc-foundation ./cmd/nsc`
- Verified the hidden command and `--remote` default through the built CLI help output.


InverseIntegral (commented):
LGTM

PR #1868, opened 2026-07-12
cli/cluster: add nsc reservation create command

annervisser:
Stacked on #1858.

Add `nsc reservation create`, with `nsc instance reserve` as an alias. It mirrors the flags of `nsc instance create` but calls the public `ReservationService.ReserveInstance` API instead of the private InstanceService.

- create returns the reservation id by default; `--wait` blocks until the reservation is fulfilled and the instance is ready, then prints the same output as `nsc instance create` (shared `printInstanceCreated`).
- `machine_type` accepts the full `[os/arch:]<cpu>x<mem>` form, with the legacy `mac/silicon` alias for `macos/arm64`; `--machine_type` stays optional.
- reuse the k3s version via `private.K3sVersion`.

Amp-Thread-ID: https://ampcode.com/threads/T-019f1880-ac56-7349-b5df-36566f7fc069

### DIFF
--- a/cmd/nsc/cmd/register.go
+++ b/cmd/nsc/cmd/register.go
@@ -35,6 +35,7 @@ func RegisterCommands(root *cobra.Command) {
 	root.AddCommand(version.NewVersionCmd())
 
 	root.AddCommand(cluster.NewBareClusterCmd("instance", false))
+	root.AddCommand(cluster.NewReservationCmd())                // nsc reservation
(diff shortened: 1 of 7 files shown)


annervisser:
Superseded by #1869. Instead of a separate `nsc reservation create` command, we're reusing this public-compute-API creation logic directly in `nsc create` behind a hidden `--use_public_api` flag. Closing.

PR #1888, opened 2026-07-20
install: retry failed downloads

n-g:
## Summary

- retry installer HTTP requests up to three times after the initial attempt
- use exponential backoff on both PowerShell and curl-based installers
- download archives fully before extraction so retries discard partial responses
- retry latest-version lookup for nsc as well as package downloads

This applies the same fresh-download behavior used by the internal artifact downloader for mid-response failures such as the HTTP/2 stream errors covered by #1886.

## Testing

- `sh -n install/install.sh`
- `sh -n install/install_nsc.sh`
- `install/install.sh --dry_run --version 0.0.546`
- `install/install_nsc.sh --dry_run --version 0.0.546`
- `git diff --check`

PowerShell is not available in the local validation environment.

### DIFF
--- a/install/install.sh
+++ b/install/install.sh
@@ -102,7 +102,8 @@ do_install() {
   fi
 
   cd ${TEMP_DIR}
-  $sh_c "curl $ci_header --fail --location --progress-bar --user-agent install.sh \"${download_uri}\" | tar -xz"
+  $sh_c "curl $ci_header --fail --location --retry 3 --retry-all-errors --progress-bar --user-agent install.sh --output ns.tar.gz \"${download_uri}\""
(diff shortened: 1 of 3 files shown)


n-g:
Validated the PR installers on real Namespace instances:

- `linux/amd64:2x4`: `install_nsc.sh` installed and ran nsc v0.0.546 from an isolated directory; `install.sh` installed and ran ns v0.0.546 from an isolated directory.
- `windows/amd64:2x4`: `install_nsc.ps1` ran under Windows PowerShell 5.1, installed `nsc.exe`, `docker-credential-nsc.exe`, and `bazel-credential-nsc.exe`, and ran nsc v0.0.546 successfully.

Both test instances were destroyed after validation.

PR #1900, opened 2026-08-06
Skip empty keep-alive responses in bazel invocation report

InverseIntegral:
## Summary

- The server now sends an initial empty `StreamInvocationReportResponse` to flush response headers before long-running report generation (namespacelabs/internal#19094).
- Skip these zero-value responses in the CLI receive loop with `proto.Equal` so they do not produce spurious `{}` output.

## Validation

- `go build ./internal/cli/cmd/cluster/...`
- `go test ./internal/cli/cmd/cluster/...`


InverseIntegral:
No longer needed. The server now uses stream.Send(nil) which flushes HTTP/2 headers without sending any DATA frame, so the client never sees a spurious message.

PR #1913, opened 2026-08-17
nsc buck2: generate buckconfig for the bazel cache and remote execution

arianvp:
Adds `nsc buck2 cache setup`, `nsc buck2 setup` and `nsc cache buck2 setup`, which provision the same backends as their bazel counterparts and emit a [buck2_re_client] buckconfig.

buck2 has no credential helper mechanism, so these always request the static auth mode and embed a bearer token in http_headers.

Unlike bazel there is no --bazelrc equivalent to point at the result. The daemon builds its RE configuration at startup from on-disk config files only (BuckConfigBasedCells::parse_with_config_args with no config args), so a file passed with --config-file never reaches [buck2_re_client]. The default destination is therefore $HOME/.buckconfig.d/50-namespace, --buckconfig redirects it, and an explicit path buck2 would not read warns. [buck2_re_client] is also absent from DaemonStartupConfig, so a running daemon keeps its old configuration; the output tells you to run `buck2 killall`.

Two further buck2 quirks the generator works around:

  - buck2 rejects every URI scheme other than grpc, dns, ipv4 and ipv6, and takes TLS from a separate `tls` key, so grpcs:// endpoints are rewritten.
  - buck2's remote execution client fails to start without engine_address, so cache-only configurations point it at the cache endpoint.

Verified against a buck2 project: with the generated config in place, `buck2 build --remote-only` executes the action remotely (remote: 1) and a rebuild after `buck2 clean` serves it from the remote action cache (100% hits). A hybrid platform races local against remote, so the output explains why cheap actions still report as local.

### DIFF
--- a/cmd/nsc/cmd/register.go
+++ b/cmd/nsc/cmd/register.go
@@ -63,12 +63,14 @@ func RegisterCommands(root *cobra.Command) {
 	root.AddCommand(cluster.NewGitCheckoutCmd())                // nsc git-checkout [hidden]
 	root.AddCommand(private.NewInternalCmd())                   // nsc internal [hidden]
 	root.AddCommand(cluster.NewBazelCmd())
+	root.AddCommand(cluster.NewBuck2Cmd())
 	root.AddCommand(cluster.NewGradleCmd())
 	root.AddCommand(cluster.NewPantsCmd())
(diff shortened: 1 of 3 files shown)


arianvp:
P.S. My ideal end-state would be that buck2 will support bazel credential helpers so we can get rid of the static token. But that's follow-up work

PR #1914, opened 2026-08-18
nsc bazel: configure build results URL

hugosantos:
## Summary

- Adds a tenant-scoped `--bes_results_url` to the bazelrc generated by `nsc bazel setup`.
- Resolves the tenant from the same token used for Bazel setup and omits the URL when build events are disabled.

## Validation

- `GOWORK=off go test ./internal/fnapi ./internal/cli/cmd/cluster`


InverseIntegral (commented):
LGTM

PR #1928, opened 2026-08-20
nsc bazel: wait for cache readiness during setup

hugosantos:
## Summary

- Wait up to one minute for Bazel storage readiness in `nsc bazel cache setup` and `nsc bazel setup`.
- Probe authenticated gRPC endpoints with `GetCapabilities` and legacy HTTP cache endpoints through `/status`.

## Validation

- `go test ./internal/cli/cmd/cluster`
- `go test -race ./internal/cli/cmd/cluster -run 'TestWaitForBazel|TestParseBazel'`
- `go build -o /tmp/nsc-bazel-readiness ./cmd/nsc`
- `git diff --check`

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,6 +3,7 @@ module namespacelabs.dev/foundation
 go 1.26.0
 
 require (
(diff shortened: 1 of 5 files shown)


InverseIntegral (commented):
LGTM

PR #1948, opened 2026-09-05
add `nsc build resolve` and `nsc build ssh`

MantasKontautas:


internal/cli/cmd/cluster/build_resolve.go
```diff
@@ -0,0 +1,159 @@
...
+			provided++
+		}
+		if *originInstance != "" {
+			provided++
+		}
+		if provided == 0 {
+			return fnerrors.BadInputError("one of --github_job, --buildkite_job, or --origin_instance is required")
+		}
+		if provided > 1 {
+			return fnerrors.BadInputError("workflow flags are mutually exclusive")
```
annervisser:
```suggestion
			return fnerrors.BadInputError("origin flags (--github_job, --buildkite_job, and --origin_instance) are mutually exclusive")
```

internal/cli/cmd/cluster/build_resolve.go
```diff
@@ -0,0 +1,159 @@
...
+	githubv1beta "namespacelabs.dev/integrations/proto/namespace/cloud/github/v1beta"
+)
+
+const buildRefAttachmentTypeURL = "namespacelabs.dev/internal/build/ref"
+
+func newBuildResolveCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "resolve",
+		Short: "Resolve container build references by origin.",
+		Args:  cobra.NoArgs,
```
annervisser: Add example for gh, bk, instance

(like https://github.com/namespacelabs/foundation/blob/8f53ba50b40d09b7dbcc14bea6023f39391d1f91/internal/cli/cmd/cluster/turborepo.go#L47-L48)

PR #1959, opened 2026-09-07
nsc: add Buck2 remote cache and execution setup

InverseIntegral:
## Key features

- Adds `nsc buck2 setup`, `nsc buck2 cache setup`, and `nsc cache buck2 setup`.
- Generates Buck2's `[buck2_re_client]` configuration in a location Buck2 reads at daemon startup.
- Supports remote execution, read-write caching, and read-only caching through the existing Bazel cache/execution APIs and readiness checks.
- Uses short-lived mTLS client certificates and direct Namespace endpoints by default, matching `nsc bazel rbe setup`; `--static` remains an explicit bearer-token fallback and `--token` implies it.
- Writes the combined client certificate/private-key PEM and generated buckconfig atomically with mode `0600`.
- Translates Namespace TLS endpoints to Buck2's address/TLS model and validates values that Buck2 would expand or parse unsafely.

## Verification

- `GIT_CONFIG_GLOBAL=/dev/null go test ./...`
- `go vet ./internal/cli/cmd/cluster ./cmd/nsc/...`
- Real Buck2 2025-12-01 remote-only build against Namespace using default mTLS: uncached action executed remotely (`remote: 1`).
- After `buck2 clean`, the same action was restored from Namespace's remote cache (`cached: 1`, `Cache hits: 100%`).
- Repeated the setup/build flow with explicit `--static`; the remote cache hit succeeded.
- Parsed the emitted combined PEM with OpenSSL and verified both the certificate chain and private key; generated config and credentials are mode `0600`.


InverseIntegral:
We've opted to use a unified command rather than a `buck2` specific one

